PR #65, opened 2019-10-07
Add data attribute to locator

flaurida:


jperl:
Game plan: fix these tests to wrap this. Then we will plan a comprehensive refactor of the locator logic to take in to account our learnings.

PR #74, opened 2019-10-13
Remove xpath as strong match

flaurida:


jperl:
Replaced by https://github.com/qawolf/qawolf/pull/77

PR #88, opened 2019-10-21
[WIP] Create GitHub package

flaurida:


packages/github/src/index.ts
```diff
@@ -0,0 +1,18 @@
+import Octokit from "@octokit/rest";
+
+const octokit = new Octokit({
+  auth: "d9e1b19607b2b07ed5c6f2e78fa5fb1bd3cabf26"
```
flaurida: this token was previously destroyed before checking in, but i did not mean to check it in. just leaving this comment here to clarify it was inactive.

PR #90, opened 2019-10-24
Make jest environment generic

flaurida:
Support running jest tests without workflow files


packages/cli/tests/runTest.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    expect(results.testResults[0].testResults).toMatchObject([
+      { status: "passed", title: "step 1" },
+      { status: "passed", title: "step 2" },
+      { status: "failed", title: "step 3" },
+      { status: "pending", title: "step 4" }
+    ]);
+
+    expect(results.testResults[0].testResults[2].failureMessages[0]).toContain(
+      "demogorgon!"
```
jperl: 😆

packages/jest-environment/tests/NodeEnvironment.test.ts
```diff
@@ -0,0 +1,6 @@
+describe("nodeEnvironment", () => {
```
jperl: should match the filename "NodeEnvironment"

PR #92, opened 2019-10-25
qawolf record

jperl:
resolves https://github.com/qawolf/qawolf/issues/93


packages/browser/src/QAWolfPage.ts
```diff
@@ -8,31 +11,66 @@ const webBundle = fs.readFileSync(
...
+  protected _events: Event[] = [];
+  private _page: DecoratedPage;
   private _requests: RequestTracker;
 
   // protect constructor to force using async create()
-  protected constructor(page: Page) {
+  protected constructor(options: CreateOptions) {
+    // decorate the page with this parent
+    const page = options.page as DecoratedPage;
+    (page as any).qawolf = this;
```
flaurida: add a type here?

packages/browser/src/QAWolfPage.ts
```diff
@@ -8,31 +11,66 @@ const webBundle = fs.readFileSync(
...
-  public static async create(page: Page, device: devices.Device) {
+  public static async create(options: CreateOptions) {
+    const { device, page } = options;
+
+    const qawolfPage = new QAWolfPage(options);
+
+    let bundle = webBundle;
+    if (options.record) {
+      // create the web Recorder and connect to this.onEvent
+      await page.exposeFunction("qaw_onEvent", (event: any) => {
```
flaurida: do we get the event type?

packages/browser/src/actions.ts
```diff
@@ -1,40 +1,52 @@
...
+    if (currentValue) {
+      // select all so we replace the text
+      // from https://github.com/GoogleChrome/puppeteer/issues/1313#issuecomment-471732011
+      await elementHandle.evaluate(() =>
+        document.execCommand("selectall", false, "")
+      );
+      await elementHandle.press("Backspace");
+    }
+
+    await elementHandle.type(strValue);
```
flaurida: this may be the airbnb issue - were you trying setting the value directly? though `type` is better i think b/c triggers onchange handlers. and now that i think about it maybe that was the missing piece for airbnb

jperl: let's sit together on this one

packages/browser/tests/actions/scroll.test.ts
```diff
@@ -0,0 +1,68 @@
+import { CONFIG } from "@qawolf/config";
```
flaurida: curious to see if horizontal scroll works too

jperl: unfortunately i don't have a good example to test with

packages/browser/src/actions.ts
jperl: note to myself: consider triggering "change/input" in js to see what happens

PR #103, opened 2019-10-29
Capture browser logs

jperl:
- name log files by the test name

### DIFF
--- a/packages/browser/src/QAWolfPage.ts
+++ b/packages/browser/src/QAWolfPage.ts
@@ -3,7 +3,7 @@ import { logger } from "@qawolf/logger";
 import { Event } from "@qawolf/types";
 import fs from "fs-extra";
 import path from "path";
-import { devices, Page } from "puppeteer";
+import { devices, JSHandle, Page } from "puppeteer";
 import { RequestTracker } from "./RequestTracker";
 
 const webBundle = fs.readFileSync(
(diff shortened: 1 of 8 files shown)


packages/browser/src/QAWolfPage.ts
```diff
@@ -77,4 +66,57 @@ export class QAWolfPage {
   public waitForRequests() {
     return this._requests.waitUntilComplete();
   }
+
+  private async captureLogs() {
+    const toString = (jsHandle: JSHandle) =>
+      jsHandle.executionContext().evaluate(obj => {
+        return obj instanceof HTMLElement
```
flaurida: maybe add a comment here to explain what this is doing? not very obvious

packages/web/src/match.ts
```diff
@@ -166,17 +166,13 @@ export const topMatch = (args: MatchArgs): Match | null => {
...
   }
 
   const equalTopMatches = matches.filter(m => m.value === matches[0].value);
   if (equalTopMatches.length > 1) {
-    console.log(
-      "no top match since all top matches are equal",
-      args,
-      equalTopMatches
-    );
+    console.log("no top match, all matches are equal", args, equalTopMatches);
```
flaurida: i think we should be super explicit in these logs - ie  no top match for `args`, all matches equal: `list of matches`

PR #104, opened 2019-10-29
setup listr for qawolf record

jperl:


packages/browser/src/QAWolfPage.ts
```diff
@@ -87,8 +87,10 @@ export class QAWolfPage {
     });
 
     const logError = (e: Error) => {
-      logger.error(
-        `page ${this._page.url().substring(0, 40)}: ${e.toString()}`
+      logger.verbose(
```
jperl: change to verbose so it does not show up when using cli by default

PR #105, opened 2019-10-29
if an invalid test is passed, prompt the user to select a test to run

jperl:


packages/cli/src/test.ts
```diff
@@ -0,0 +1,44 @@
+import { snakeCase } from "lodash";
+import { basename } from "path";
+import { runTest } from "./runTest";
+import { readdir } from "fs-extra";
```
flaurida: abc

packages/cli/src/index.ts
```diff
@@ -30,29 +27,11 @@ recordCommand.action(async (urlArgument, optionalName, cmd) => {
   await record(url, name, cmd.events);
 });
 
-program
```
jperl: removed for now because we are not supporting this. when we do we should gracefully handle incorrect names like in test.

PR #110, opened 2019-10-31
waitUntil and getElementProperty helpers

flaurida:


packages/browser/src/assertions.ts
```diff
@@ -26,3 +26,31 @@ export const hasText = async (
...
+
+export const getProperty = async (
+  page: Page,
+  { property, selector }: GetPropertyArgs,
+  options?: AssertOptions
+): Promise<string | null | undefined> => {
+  const result = await retryExecutionError(async () => {
+    const timeoutMs = (options || {}).timeoutMs || CONFIG.findTimeoutMs;
+
+    const getElementHandleFn = async () => {
```
jperl: I found it confusing you defined this function here, instead of just directly inside the waitFor

packages/browser/tests/assertions.test.ts
```diff
@@ -1,34 +1,78 @@
...
+    });
+    expect(placeholder).toBeUndefined();
+  });
+
+  it("returns null if no elements match selector", async () => {
+    const tagName = await getProperty(page, {
+      selector: "#wrongId",
+      property: "tagName"
+    });
+    expect(tagName).toBeNull();
```
jperl: I would set a short timeout for these null tests

PR #119, opened 2019-11-05
⌨️ Support keyboard events ✏️Support content editables 🚫🤖🖱️do not record non-user scrolls

### LINKED ISSUE #60: ✏️Support content editables

### LINKED ISSUE #118: ⌨️ Support keyboard events
After testing against various sites (google flights/hipmunk/united) it is clear we need to support Enter and Tab -- which are both ways to register the input. Ex. When you just set the input value "JFK" it does not work. But if you `Enter`, `Tab`, or click the airport dropdown item it registers.

We could try to hack this by simulating a change event or something like this. However our philosophy (soon to be added to the readme) is to match the user behavior as much as possible to prevent a leaky abstraction. That way we do not make assumptions about how the site is implemented.

This will also allow us to support content editables and hotkeys.

To Do:

- [x] Capture `keydown` and `keyup` events and use those for building type steps, instead of `input` events
- [x] Support key sequences (CMD+A, etc) with syntax `↓MetaLeft↓KeyA↑MetaLeft↑KeyA`
- [x] Test everything manually
- [x] Separate enter and tab steps to make it easy to parameterize
- [x] Convert values w/o arrow prefix to ^ using [USKeyboardLayout](https://github.com/GoogleChrome/puppeteer/blob/master/lib/USKeyboardLayout.js)
- [x] Add back select support by tracking input events
- [x] Replace ctrl+v keystroke with paste value
- [x] Add tests for everything

### LINKED ISSUE #125: 🚫🤖🖱️do not record non-user scrolls
addEventListener for [wheel](https://stackoverflow.com/questions/7035896/detect-whether-scroll-event-was-created-by-user#comment98810714_7210072) instead of `scroll`

jperl:
Resolves #60
Resolves #118
Resolves #125 

- separates the concepts of EventName and Action


packages/browser/src/actions.ts
```diff
@@ -1,38 +1,31 @@
-import { QAWolfWeb } from "@qawolf/web";
+import { logger } from "@qawolf/logger";
+import { QAWolfWeb, sleep } from "@qawolf/web";
```
jperl: abc

packages/browser/src/actions.ts
```diff
@@ -1,38 +1,31 @@
...
+    element => (element as HTMLInputElement).value
+  );
 
-    const currentValue = await elementHandle.evaluate(
-      element => (element as HTMLInputElement).value
+  if (currentValue) {
+    // select all so we replace the text
+    // from https://github.com/GoogleChrome/puppeteer/issues/1313#issuecomment-471732011
+    await elementHandle.evaluate(() =>
+      document.execCommand("selectall", false, "")
```
jperl: why do this vs setting value to `""`?

jperl: that does not mimic user behavior and may not work on some sites

packages/browser/src/keyboard.ts
```diff
@@ -0,0 +1,128 @@
...
+
+// A Stroke is a single keydown or keyup
+export type Stroke = {
+  code: string;
+  index: number;
+  prefix: "↓" | "↑";
+};
+
+// organizes the KeyDefinitions from USKeyboardLayout
+const keyToDefinition: { [key: string]: KeyDefinition } = {};
```
flaurida: what about non-US keyboards? will this still work?

jperl: Probably not, but puppeteer does not have a non-USKeyboardLayout version. https://github.com/GoogleChrome/puppeteer/tree/master/lib

flaurida: OK, let's see if it works 🤷‍♀ apparently US layout pretty common

packages/browser/src/actions.ts
jperl: added a comment clarifying this

packages/browser/tests/web/Recorder.test.ts
```diff
@@ -27,90 +27,145 @@ describe("Recorder", () => {
...
     // close the browser to ensure events are transmitted
     await browser.close();
 
-    const events = page.qawolf.events.filter(e => e.isTrusted);
-    const lastEvent = events[events.length - 1] as InputEvent;
-    expect(lastEvent.target.xpath).toEqual("//*[@id='password']");
-    expect(lastEvent.value).toEqual("secret");
+    const events = page.qawolf.events;
+    expect(events[0].target.xpath).toEqual("//*[@id='password']");
+    expect((events[0] as InputEvent).value).toEqual("secret");
```
jperl: wrong cast

packages/browser/src/keyboard.ts
jperl: It works. It just probably won't work for non-US layouts :/.

packages/web/src/Recorder.ts
```diff
@@ -19,12 +19,28 @@ export class Recorder {
...
+
   private recordEvent<K extends keyof DocumentEventMap>(
     eventName: K,
-    handler: (ev: DocumentEventMap[K]) => types.Event
+    handler: (ev: DocumentEventMap[K]) => types.Event | undefined
   ) {
-    const listener = (ev: DocumentEventMap[K]) => {
+    this.listen(eventName, (ev: DocumentEventMap[K]) => {
       const event = handler(ev);
+      if (!event) return;
```
flaurida: should we not return `null` instead to be more explicit?

jperl: nah

jperl: I'd rather not return anything

PR #136, opened 2019-11-11
Wait for option with specified value in select

### LINKED ISSUE #133: ✋Wait for select value
Auto-wait for the select value before proceeding

flaurida:
* closes #133

### DIFF
--- a/packages/browser/src/actions.ts
+++ b/packages/browser/src/actions.ts
@@ -59,9 +59,10 @@ export const select = async (
 ): Promise<void> => {
   logger.verbose("actions.select");
   // ensure option with desired value is loaded before selecting
-  await waitUntil(async () => {
-    const isOptionAvailable = await elementHandle.evaluate(
-      (element: HTMLSelectElement, value: string | null) => {
+  await elementHandle.evaluate(
+    (element: HTMLSelectElement, value: string | null, timeoutMs: number) => {
+      const qawolf: QAWolfWeb = (window as any).qawolf;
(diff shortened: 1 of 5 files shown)


packages/browser/src/actions.ts
```diff
@@ -53,9 +54,28 @@ export const scroll = async (
 
 export const select = async (
   elementHandle: ElementHandle,
-  value: string | null
+  value: string | null,
+  timeoutMs?: number
 ): Promise<void> => {
   logger.verbose("actions.select");
+  // ensure option with desired value is loaded before selecting
+  await waitUntil(async () => {
```
jperl: Our waiting logic should all happen in the browser if possible. Can you move this to a QAWolfWeb function and use it in the action? Similar to scroll.

packages/browser/src/actions.ts
```diff
@@ -53,9 +54,28 @@ export const scroll = async (
...
   elementHandle: ElementHandle,
-  value: string | null
+  value: string | null,
+  timeoutMs?: number
 ): Promise<void> => {
   logger.verbose("actions.select");
+  // ensure option with desired value is loaded before selecting
+  await elementHandle.evaluate(
+    (element: HTMLSelectElement, value: string | null, timeoutMs: number) => {
+      const qawolf: QAWolfWeb = (window as any).qawolf;
```
jperl: I think we might want to keep the web library code inside the web package to keep things organized.

Perhaps a new file select, with `qawolf.select.waitForOption` with this as a helper>

PR #147, opened 2019-11-11
run tests on fork PRs

jperl:


jperl:
Cannot do this... we have docker secrets :(

jperl:
We could on `pull_request` assigned. But it's better to have to clone the branch first and do a deep dive before running the tests to make sure the PR does not expose secrets.

jperl:
Wait never mind the docker push is a separate workflow!! :)

PR #153, opened 2019-11-11
👴Clickable ancestors

### LINKED ISSUE #129: 👴Clickable ancestors
- [x] Crawl to the top clickable ancestors (short-circuit at data attribute) and use that as the target 
- [x] Test w/ 🗞️Data attribute example for NY Times -- QAW_DATA_ATTRIBUTE: data-testid

jperl:
Closes #129

Find highest clickable ancestor (stopping at closest w/ data attribute) and use that as the click target.

- [x] explain why we want to do this in a comment (svgs, icons, etc)
- [x] clean up API -- getComputedStyle is used in multiple places/should isClickable go in element?/etc...
- [x] test against many sites
- [x] add unit tests for findClickableAncestor

### DIFF
--- a/packages/web/src/element.ts
+++ b/packages/web/src/element.ts
@@ -179,10 +179,6 @@ export const isVisible = (
 ): boolean => {
   if (element.offsetWidth <= 0 || element.offsetHeight <= 0) return false;
 
-  // make sure the element is large enough to interact with
-  const rect = element.getBoundingClientRect();
-  if (rect.width <= 0 || rect.height <= 0) return false;
-
   if (computedStyle && computedStyle.visibility === "hidden") return false;
 
   return true;
--- a/packages/web/src/element.ts
+++ b/packages/web/src/element.ts
@@ -163,3 +163,27 @@ export const getDescriptor = (
     xpath
   };
 };
+
+export const isClickable = (
+  element: HTMLElement,
+  computedStyle: CSSStyleDeclaration
+) => {
+  // assume it is clickable if the cursor is a pointer
+  const clickable = computedStyle.cursor === "pointer";
+  return clickable && isVisible(element, computedStyle);
+};
+
+export const isVisible = (
+  element: HTMLElement,
+  computedStyle?: CSSStyleDeclaration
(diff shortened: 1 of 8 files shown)


jperl:
Relates to https://github.com/qawolf/examples/pull/1

packages/web/src/find.ts
```diff
@@ -78,8 +115,9 @@ export const findElement = async ({
   if (target.xpath === "/html" || target.xpath === "/html/body")
     return findElementByXpath(target.xpath);
 
+  console.log("waiting for top strong match", target);
```
jperl: moved this outside of the waitFor so we don't have a bunch of repetitive logs

packages/web/src/element.ts
```diff
@@ -163,3 +163,27 @@ export const getDescriptor = (
...
+};
+
+export const isVisible = (
+  element: HTMLElement,
+  computedStyle?: CSSStyleDeclaration
+): boolean => {
+  if (element.offsetWidth <= 0 || element.offsetHeight <= 0) return false;
+
+  // make sure the element is large enough to interact with
+  const rect = element.getBoundingClientRect();
```
flaurida: don't use this method it is very slow and non performant

packages/web/src/element.ts
```diff
@@ -163,3 +163,27 @@ export const getDescriptor = (
...
+  // assume it is clickable if the cursor is a pointer
+  const clickable = computedStyle.cursor === "pointer";
+  return clickable && isVisible(element, computedStyle);
+};
+
+export const isVisible = (
+  element: HTMLElement,
+  computedStyle?: CSSStyleDeclaration
+): boolean => {
+  if (element.offsetWidth <= 0 || element.offsetHeight <= 0) return false;
```
flaurida: make sure using right `width` method: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements

jperl: I didn't write that code :0. But it seems right to me.

PR #166, opened 2019-11-13
findClickableAncestor: stop at expected clickable element types

jperl:


packages/web/src/find.ts
```diff
@@ -68,7 +68,18 @@ export const findClickableAncestor = (
       return ancestor;
     }
 
-    // short-circuit when parent is not clickable
+    // short-circuit when we encounter a common clickable element type
+    // there may be a parent that is still clickable but does something else
+    if (
```
flaurida: just use input in general for now. also transform to lower case in case tag name isn't upper case (not 100% sure but at least safer imo)

PR #167, opened 2019-11-13
qawolf docker: remove entrypoint & support parallel qawolf calls w/ multiple xvfb displays

jperl:
- [x] remove entrypoint to support azure pipelines https://github.com/qawolf/qawolf/issues/165 
- [x] start xvfb display inside qawolf alias, since there is no entrypoint -- this will support multiple calls/displays
- [x] setup in https://github.com/qawolf/examples/pull/4 to make sure it works


jperl:
TODO fix qawolf-npm-test not passing errors back to stderr

Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN mkdir -p /etc/opt/chrome/policies/managed && echo "{ \"CommandLineFlagSecuri
 
 # Copy everything and install dependencies in a static location that is not WORKDIR
 # WORKDIR will be replaced by github action volume
-ENV QAWOLF_DIR "/root/qawolf"
+ENV QAWOLF_DIR "/qawolf/src"
```
jperl: Changed this from /root/qawolf since that caused issues w/ azure pipelines

PR #171, opened 2019-11-18
cli: add "qawolf azure" and "qawolf circleci" commands

jperl:
- remove unused @qawolf/github package for now


packages/cli/src/ci.ts
```diff
@@ -0,0 +1,33 @@
+import { copy, pathExists } from "fs-extra";
+import { prompt } from "inquirer";
+import path from "path";
+
+type CiProvider = "azure" | "circleci" | "github";
```
flaurida: technically I think should be `CIProvider`

packages/cli/static/azure.yml
```diff
@@ -0,0 +1,31 @@
+trigger:
```
flaurida: maybe add a comment showing how to change?

PR #173, opened 2019-11-19
chore: minor code refactoring

jperl:
continuation of https://github.com/qawolf/qawolf/pull/170

### DIFF
--- a/packages/runner/src/Runner.ts
+++ b/packages/runner/src/Runner.ts
@@ -162,7 +162,7 @@ export class Runner {
       // do not focus or clear for Enter or Tab
       const shouldFocusClear =
         !typeValue ||
-        (!typeValue.includes("↓Enter") && !typeValue.includes("↓Tab"));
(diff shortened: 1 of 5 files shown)


packages/runner/src/Runner.ts
```diff
@@ -162,7 +162,7 @@ export class Runner {
       // do not focus or clear for Enter or Tab
       const shouldFocusClear =
         !typeValue ||
-        (typeValue.indexOf("↓Enter") !== 0 && typeValue.indexOf("↓Tab") !== 0);
+        !(typeValue.startsWith("↓Enter") || typeValue.startsWith("↓Tab"));
```
jamesgeorge007: Maybe :thinking: 
```js
!(["↓Enter", "↓Tab"].some(key => typeValue.startsWith(key) ));
```

jperl: 🤔

jamesgeorge007: Never mind, the present approach is pretty straight forward it seems.

PR #211, opened 2019-12-02
do not choose non-common clickable element types

jperl:
If a target is the descendant of a button/a/input we want to click the parent -- otherwise choose the target. This is to avoid the issue where a parent div does something different. Eg dropdown divs.


jperl:
Relates to #129 #153

PR #217, opened 2019-12-03
Use fixed qawolf version in generated CI files

### LINKED ISSUE #210: 🏷️Use fixed version in `qawolf [ci]` templates
- [x] Use [version](https://github.com/qawolf/qawolf/blob/master/packages/cli/src/index.ts#L12) from `package.json` in template w/ handlebars
- [x] Remove github action, use docker directly see https://github.com/qawolf/nytimes-example/commit/4340ab4dcab7852751f5d3bae869daee9972e970

flaurida:
* closes #210


packages/cli/static/github.hbs
```diff
@@ -26,7 +32,7 @@ jobs:
         #   QAW_LOGIN_0: "my_username"
         #   # you can also use GitHub secrets for environment variables
         #   # https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets
-        #   QAW_LOGIN_1: ${{ secrets.PASSWORD }}
+        #   QAW_LOGIN_1: $\{{ secrets.PASSWORD }}
```
flaurida: need to escape the `{{` so handlebars doesn't get confused

jperl:
delete https://github.com/qawolf/qawolf/blob/master/action.yml since it is no longer used

PR #230, opened 2019-12-13
findText: find the best matching element and make it case sensitive

jperl:
- findCss & findText: store last find on window.qaw_find for easy debugging
- hasText: add logging
- hasText: add test for case sensitivity

### DIFF
--- a/packages/browser/src/find/hasText.ts
+++ b/packages/browser/src/find/hasText.ts
@@ -16,7 +16,7 @@ export const hasText = async (
       return qawolf.find.hasText(text, timeoutMs);
     },
     text,
-    findOptions.timeoutMs || 0
+    findOptions.timeoutMs!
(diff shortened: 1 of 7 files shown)


packages/browser/src/find/hasText.ts
```diff
@@ -16,7 +16,7 @@ export const hasText = async (
       return qawolf.find.hasText(text, timeoutMs);
     },
     text,
-    findOptions.timeoutMs || 0
+    findOptions.timeoutMs!
```
flaurida: how do we know will be defined? if it will seems like the type should be `number` and the `!` unnecessary?

jperl: It gets defined in getFindElementOptions if it is nil, but I will put it back for clarity.

PR #234, opened 2019-12-13
fix queryElements type to be Element

jperl:
- it was mistyped to be HTMLElement[] leading to bugs


packages/web/src/find/findText.ts
```diff
@@ -21,6 +20,9 @@ export const findText = async (
       for (let i = 0; i < elements.length; i++) {
         const element = elements[i] as HTMLElement;
 
+        // skip non-HTML elements
+        if (!element.innerText) continue;
```
jperl: this fixes an undefined exception with findText

PR #237, opened 2019-12-16
Escape single quote characters in test name

flaurida:
Resolves https://github.com/qawolf/qawolf/issues/236

### DIFF
--- a/packages/build-code/src/formatIt.ts
+++ b/packages/build-code/src/formatIt.ts
@@ -23,6 +23,7 @@ export const formatDescription = (step: Step): string => {
   }
 
   if (description.length) {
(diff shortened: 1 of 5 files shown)


packages/build-code/src/formatIt.ts
```diff
@@ -22,7 +22,10 @@ export const formatDescription = (step: Step): string => {
     description = `${description.substring(0, 40)}...`;
   }
 
-  if (description.length) return ` "${description}"`;
+  if (description.length) {
+    // strip single quotes so they don't mess up formatting
```
jperl: // strip single quotes so the description is valid javascript

PR #244, opened 2019-12-19
⌨️Improve recorder hotkey support

### LINKED ISSUE #242: ⌨️Improve recorder hotkey support
Relates to #240 https://github.com/qawolf/qawolf/pull/241

Test on github shortcuts `g` + `i`

### LINKED ISSUE #243: Improve hasText documentation
* `hasText` actually waits until text is visible, rather than returning true or false
* Also, "timeoutMs <?number> maximum time to wait for a page to open. Defaults to 5000." I believe is actually the amount of time to wait for the text to appear before giving up
* An alternative is making `hasText` simply return the boolean, and then add an `assertText` helper that waits for text to appear and then fails if it doesn't before timeout

jperl:
Resolves #242

- buildTypeSteps use simplified [serializeKeyEvents](https://github.com/qawolf/qawolf/issues/242#issuecomment-567561669) instead of simplifyStrokes & buildPendingStrokesStep
- simplify and make serializeStrokes and deserializeStrokes consistent
- change queryDataElements to only select based on data attribute and not care about action type. If someone specifies a data attribute we should not be picky
- findHtml: find html & body always (including for type actions) --- ex. to support hotkeys
- type: add options.delayMs default to 300ms for keyboard and 0ms for sendCharacter
- buildTypeSteps: add snapshot test with github shortcut fixture
- remove QAW_KEY_DELAY_MS env variable
- fix hasText options.timeoutMs documentation. resolves #243


packages/browser/src/actions/typeElement.ts
```diff
@@ -33,14 +34,15 @@ export const typeElement = async (
     if (stroke.type === "↓") {
       logger.debug(`keyboard.down("${stroke.value}")`);
       await page.keyboard.down(stroke.value);
+      await sleep(isNil(options.delayMs) ? 300 : 0);
```
flaurida: nit: if `sleep` is passed `0` should return, not return a new promise that resolves in `0` seconds

jperl: k

packages/web/src/wait.ts
```diff
@@ -1,4 +1,6 @@
 export const sleep = (milliseconds: number): Promise<void> => {
+  if (milliseconds <= 0) return;
```
flaurida: 😍

PR #246, opened 2019-12-20
Fix incorrect argument description in hasText documentation

### LINKED ISSUE #243: Improve hasText documentation

flaurida:
* closes #243

### DIFF
--- a/docs/docs/api.md
+++ b/docs/docs/api.md
@@ -262,8 +262,7 @@ const page = await browser.goto("https://youtube.com", {
 - `text` <[string]> the text to find on the page.
 - `options` <[FindPageOptions]> find the page with these options.
   - `page` <?[number]> the index of the page to use in order of creation, starting with 0. defaults to the last used page.
-  - `timeoutMs` <?[number]> maximum time to wait for a page to open. Defaults to `5000`.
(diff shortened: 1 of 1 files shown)


docs/docs/api.md
```diff
@@ -262,8 +262,7 @@ const page = await browser.goto("https://youtube.com", {
 - `text` <[string]> the text to find on the page.
 - `options` <[FindPageOptions]> find the page with these options.
   - `page` <?[number]> the index of the page to use in order of creation, starting with 0. defaults to the last used page.
-  - `timeoutMs` <?[number]> maximum time to wait for a page to open. Defaults to `5000`.
-  - `waitForRequests` <?[boolean]> wait until the page completes all network requests (limited to 10s per request). Defaults to `true`.
```
jperl: waitForRequests is still a valid option why did you remove it. also it should be `Defaults to [QAW_TIMEOUT_MS](#qaw_timeout_ms).`

jperl:
I will fix this in my PR

PR #293, opened 2020-01-03
Rename qawolf record -> qawolf create

### LINKED ISSUE #259: 🧞Rename cli record to create
Record is confusing because we also record a video. The outcome of the command is to create a browser script or test. 

- [ ] rename `qawolf record` -> `qawolf create`
- [ ] recommend `npx qawolf test` instead of `npx qawolf test [name]`

flaurida:
* Resolves #259

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -19,7 +19,7 @@ program
   .command("create <url> [name]")
   .option("--debug", "save events and workflow json for debugging")
   .option("-d, --device <device>", "emulate using a puppeteer.device")
-  .option("-s, --script", "save a script instead of a test")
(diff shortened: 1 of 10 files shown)


packages/cli/src/index.ts
```diff
@@ -16,17 +16,17 @@ program.version(version);
 program.usage("<command> [options]");
 
 program
-  .command("record <url> [name]")
+  .command("create <url> [name]")
   .option("--debug", "save events and workflow json for debugging")
   .option("-d, --device <device>", "emulate using a puppeteer.device")
   .option("-s, --script", "save a script instead of a test")
```
jperl: create a script instead of a test

Issue #292, opened 2020-01-02
🔮Inline css selector for find attribute & inline page


jperl:
Benefits
- The selector with the find attribute is inlined in code, making it clear to the developer it is used
- The page option is inlined in code, making it clearer
- The developer will not need to specify QAW_FIND_ATTRIBUTE when running tests
- (Combined with #295) the selector is simplified to an object with one key
- findHtml/queryElements code is simplified, we do not need separate logic for when there is a data attribute

To Do

- [x] in stepToSelector: choose html or css, depending on whether the node has the data attribute. Add tests
- [x] change`formatMethod(step.action, step.index, step.value)` -> `formatMethod(step, previousStep?)`. 
- [x] create `formatSelector` and use it in `formatMethod`. Use `stepToSelector` and inline selector if it is css. Add tests.
- [x] move page from `Selector` -> `FindElementOptions` and inline it in the test. Make `FindElementOptions` extend `FindPageOptions`. In `QAWolfBrowser` change `selector.page` -> `options.page`
- [x] in `formatMethod` inline options if page is different than the previous step: `click(selectors[0], { page: 1 })`
- [x] remove queryElements, rename queryActionElements -> queryElements. Remove queryDataElements and related tests. Remove dataAttribute from `QueryElementsOptions`
- [x] remove dataAttribute from FindElementOptions, matchElements, matchDocSelector, and related tests
- [x] rename QAW_DATA_ATTRIBUTE -> QAW_CSS_ATTRIBUTE
- [x] make sure docs reflect these changes (how it works/CLI env variables/API)

jperl:
We should also probably rename it from data attribute -> css attribute, since theoretically you could use any css attribute (id,etc).

`QAW_DATA_ATTRIBUTE` -> `QAW_CSS_ATTRIBUTE`

jperl:
```
const { launch } = require("qawolf");
const selectors = require("../selectors/search");
describe("search", () => {
  let browser;
  beforeAll(async () => {
    browser = await launch({ url: "https://nytimes.com/" });
  });
  afterAll(() => browser.close());
  it('can click "SEARCH" button', async () => {
    await browser.click(selectors[0]);
  });
  it('can type into "query" input', async () => {
    await browser.type({ css: "[data-testid='search-input']" }, "github");
  });
  it("can Enter", async () => {
    await browser.type({ css: "[data-testid='search-input']" }, "↓Enter↑Enter");
  });
  it("can scroll", async () => {
[...]

PR #295, opened 2020-01-03
change Selector to be a union type

jperl:
- rename serializeStep -> stepToSelector -- remove StepSerialized instead output Selector
- combine htmlSelectorToDocSelector into deserializeDocSelector
- move action from Selector -> FindElementOptions

Prep work for #292

### DIFF
--- a/packages/browser/src/find/find.ts
+++ b/packages/browser/src/find/find.ts
@@ -1,5 +1,11 @@
 import { logger } from "@qawolf/logger";
-import { FindElementOptions, Selector } from "@qawolf/types";
+import {
+  CssSelector,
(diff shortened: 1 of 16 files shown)


jperl:
Things I noticed

- findCss does not use queryActionElements, should it?
- in #292 we should move page from `Selector` -> `FindElementOptions` and inline it in the test. Maybe `FindElementOptions` should extend `FindPageOptions`

jperl:
- ~~findHtml and findText use queryActionElements but findCss does not should it?~~ nah if you are specifying the css selector than I would expect that to run, not some modified version of it

PR #312, opened 2020-01-10
Fix 'Invalid command "create"' error in README.md

kanziw:
I have referenced [this documentation](https://docs.qawolf.com/docs/quick_start#-create-a-browser-test).

image


jperl:
The next version, which has not been released it will be renamed to create. But I guess we should not have updated the readme yet. Thank you 🙏

flaurida:
Thanks @kanziw 🙌

PR #333, opened 2020-01-19
feat(cli): Add path option

isy:
resolves #332

Added path option to cli.

QA Wolf is a great product. Thanks☺️

PR #335, opened 2020-01-19
feat(cli): Add path option

jperl:
Tweaks for #333

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -9,7 +9,7 @@ import { saveCiTemplate } from "./ci";
 import { create } from "./create";
 import { howl } from "./howl";
 import { runJest } from "./runJest";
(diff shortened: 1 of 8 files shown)


packages/cli/src/index.ts
```diff
@@ -40,17 +41,23 @@ program
...
 
 program
   .command("test")
+  .option("-p, --path <path>", "path to test code")
   .description("run a test with Jest")
   .allowUnknownOption(true)
-  .action(() => {
-    const code = runJest(process.argv.slice(3));
+  .action(cmd => {
+    const args = omitArgs(process.argv.slice(3), ["-p", "--path"]);
```
isy: Thanks for tweaking @jperl 😀

This code might solve with commander.js v5
https://github.com/tj/commander.js/pull/1138

PR #370, opened 2020-02-01
Docs revamp

flaurida:
* docs part of #366


docs/docs/add_assertions.md
```diff
@@ -1,4 +1,102 @@
...
+title: 💪 Add Assertions
 ---
+
+In this guide we show you how to add assertions to your tests. We assume that you know how to [create a test](create_a_test).
+
+## TL;DR
+
+- [Use QA Wolf helpers](#use-qa-wolf-helpers) to create assertions
+- [Use the Playwright API](#use-the-playwright-api) to create assertions
+- The [interactive REPL](use_the_repl) lets you try out assertions while creating tests
```
jperl: For consistency

- [Use the interactive REPL] to ...

docs/src/hotjar/index.js
```diff
@@ -0,0 +1,25 @@
+module.exports = function() {
+  return {
+    name: "hotjar",
+
+    injectHtmlTags() {
+      return {
+        headTags: [
```
jperl: return null if there is no hotjar env variable

docs/docs/privacy_policy.md
```diff
@@ -0,0 +1,8 @@
+---
+id: privacy_policy
+title: Privacy Policy
+---
+
+We use Hotjar in order to better understand our users’ needs and to optimize this service and experience. Hotjar is a technology service that helps us better understand our users’ experience (for example: how much time they spend on which pages, which links they choose to click, and what users do and don’t like) and this enables us to build and maintain our service with user feedback. Hotjar uses cookies and other technologies to collect data on our users’ behavior and their devices. This includes a device's IP address (processed during your session and stored in a de-identified form), device screen size, device type (unique device identifiers), browser information, geographic location (country only), and the preferred language used to display our website. Hotjar stores this information on our behalf in a pseudonymized user profile. Hotjar is contractually forbidden to sell any of the data collected on our behalf.
```
jperl: You should make it clear this is only on our docs we use hotjar.

jperl: And that the reason we use it is to see what parts of our documentation people use so we can invest in improving them.

PR #373, opened 2020-02-03
repl() to open repl

### LINKED ISSUE #127: ✋repl() to pause and open REPL
Problem: Make it easy to fix a part of a test.

Solution: 

- `await repl()` that will open a REPL
- `resume()` to close REPL and continue the rest of the file

jperl:
Resolves #127

To Do:

- [x] on repl close, call resume
- [x] test it works on `node script.js`
- [x] setup basic reporter
- [x] `npx qawolf test --repl` that sets up reporter
- [x] make `repl` a global
- [x] add tests

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -53,11 +53,8 @@ program
   .option("-p, --path <path>", "path to test code")
   .option("--all-browsers", "run tests on chromium, firefox, and webkit")
   .option("--chromium", "run tests on chromium")
-  .option(
-    "--repl",
-    "override the default Jest reporter since it breaks the repl"
-  )
   .option("--firefox", "run tests on firefox")
+  .option("--repl", "open a REPL when repl() is called")
   .option("--webkit", "run tests on webkit")
   .description("run a test with Jest")
   .allowUnknownOption(true)
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -53,6 +53,10 @@ program
   .option("-p, --path <path>", "path to test code")
   .option("--all-browsers", "run tests on chromium, firefox, and webkit")
   .option("--chromium", "run tests on chromium")
+  .option(
+    "--repl",
+    "override the default Jest reporter since it breaks the repl"
+  )
   .option("--firefox", "run tests on firefox")
   .option("--webkit", "run tests on webkit")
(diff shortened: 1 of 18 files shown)


packages/cli/src/index.ts
```diff
@@ -53,6 +53,10 @@ program
   .option("-p, --path <path>", "path to test code")
   .option("--all-browsers", "run tests on chromium, firefox, and webkit")
   .option("--chromium", "run tests on chromium")
+  .option(
+    "--repl",
+    "override the default Jest reporter since it breaks the repl"
```
flaurida: this doesn't seem like what the command really does? something more like "use interactive repl to run code?"

flaurida: (does from a user perspective)

packages/repl/src/repl.ts
```diff
@@ -0,0 +1,42 @@
...
+import { bold } from "kleur";
+import { start, REPLServer } from "repl";
+const { addAwaitOutsideToReplServer } = require("await-outside");
+
+export const repl = (
+  context: any = {},
+  onReplCreated?: (replServer: REPLServer) => void
+) => {
+  logger.debug("repl: start");
+  console.log(bold().yellow("Type .exit or resume() to close the repl"));
```
flaurida: do `.exit` and `resume()` do the same thing? I doubt it...let's clarify this message

packages/repl/src/repl.ts
```diff
@@ -0,0 +1,42 @@
+import { logger } from "@qawolf/logger";
+import { bold } from "kleur";
+import { start, REPLServer } from "repl";
+const { addAwaitOutsideToReplServer } = require("await-outside");
```
flaurida: import doesn't work here? just confirming

packages/repl/src/repl.ts
jperl: Good point. Removing resume()

packages/repl/src/repl.ts
jperl: They do not have typescript support, so I am just leaving as is. Will add a comment.

jperl: I'll add types 👍

jperl: I added types

PR #380, opened 2020-02-04
👷Create code in realtime

### LINKED ISSUE #361: ✏️Generate editable code in realtime
- Generate code in realtime and allow it to be edited
- Write `// 🐺CREATE CODE HERE` and allow it to be moved

### LINKED ISSUE #376: ✏️Create in an existing file
Allow passing an existing file to npx qawolf create to generate more code.

Where `await create();` is called, new steps will be added. If `await create();` is not specified, an error will be thrown.

Relates to #298 #405

### LINKED ISSUE #379: ➡️Rename context back to browser
Just the variable in code, not the class (until playwright does). 

`context = launch()` -> `browser = launch()`

- qawolf only uses one context per browser by default
- conceptually less confusing
- less confusing for existing users, since their code will not change
- playwright might be doing the same https://github.com/microsoft/playwright/issues/816

jperl:
Resolves #361
Resolves #379 

To Do

- [x] architect it
- [x] stream events from browser
- [x] Step.isFinal
- [x] buildCode -> buildInitialCode
- [x] setup CodeWriter to create file when not editing
- [x] Use CodeWriter in CreateCodeCLI
- [x] CodeUpdater.prepareSteps w/ tests
- [x] remove handlebars dependency from build-code. we don't generate html, so can use string interpolation
- [x] context -> browser
- [x] CodeUpdater.updateCode w/ tests
- [x] match indentation
- [x] refactor & add lots of tests
- [x] finish TODOS


jperl:
CodeBuilder outline

```
const CREATE_SYMBOL = "🐺 CREATE CODE HERE";

class CodeBuilder {
  initialize() {
    // create file from template
    // or add the create symbol
  }

  finalize() {
    // remove the create symbol
  }

  prepareSteps() {
    // const workflow = buildWorkflow();
    // workflow.steps.filter(step => step.isFinal).push(this._pendingSteps);
  }

  patchSteps(code: string) {
    // load file, find create symbol
    // replace the appended line
    // with the code for each pending step
    // log error if create symbol is not found (debounced)
  }
}

// CreateCommand
// if (!updateFile) {
//   factory.create(path);
// }
// this._browser.listen(
//   "event",
//   debounce(event => {
//     builder.prepareSteps();
//   })
// );
// run on an interval
// write(builder.patchSteps(loadFile()))
```

packages/build-code/src/buildMethod.ts
```diff
@@ -0,0 +1,42 @@
+import { ScrollValue, Step } from "@qawolf/types";
```
jperl: Didn't actually change much in this file, just added `await` and `;` to methods.

lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.9.0-alpha.7"
```
flaurida: version going down?

jperl: Don't worry about this, since beta.0 is next.

packages/build-code/src/StepBuilder.ts
```diff
@@ -0,0 +1,64 @@
...
+  private _steps: Step[] = [];
+
+  // used for code with existing steps
+  private _startIndex: number;
+
+  public constructor({ startIndex }: ConstructorOptions = {}) {
+    this._startIndex = startIndex || 0;
+  }
+
+  private _buildSteps({ canChange }: { canChange: boolean }) {
```
flaurida: why take an object as an arg if it's just one param?

packages/build-code/src/StepBuilder.ts
```diff
@@ -0,0 +1,64 @@
...
+    this._steps.push(...newSteps);
+  }
+
+  public finalize() {
+    this._finalized = true;
+
+    this._buildSteps({
+      // there are no new events coming in
+      // so we know the steps that canChange will not
+      canChange: true
```
flaurida: why is `canChange: true` if the steps won't change?

jperl: This is trying to say. 

`// Build steps even if they can change. We know they will not change, since there are no events coming in.`

Perhaps I should rename it to `allowCanChangeSteps`? And update the comment to ^

packages/build-code/src/StepBuilder.ts
jperl: Ok I `canChange` that. Get it 😛

jperl: Actually after re-reviewing, I use an object so it is explicit in where I call it what option I am passing. I went back and forth on this, but I think it is important.

packages/qawolf/src/postInstallMessage.js
```diff
@@ -1,8 +1,11 @@
...
+  // setTimeout so it logs after everything else, hopefully
+  console.log(bold().blue("🐺  QA Wolf successfully installed!"));
 
-console.log(
-  bold().blue("✨  Create your first test by running:"),
-  "npx qawolf create <url> [name]\n"
-);
+  console.log(
+    bold().blue("✨  Create your first test by running:"),
+    "npx qawolf create <url> [name]\n"
```
flaurida: change `<url>` -> `url` to be consistent with CLI docs

PR #381, opened 2020-02-04
rename qawolf.log -> server.log

jperl:


jperl:
Nah JK. We want qawolf logs to mainly be for internal use.

PR #387, opened 2020-02-08
Fix multi-page test generation

jperl:
- include @qawolf/fixture as @qawolf/test package (since the other is versioned too far ahead)
- remove unused buildWorkflow
- rename StepBuilder.getEvents -> events for consistency
- CodeCreator: save debug file


packages/cli/src/CreateCodeCLI.ts
```diff
@@ -94,7 +98,7 @@ export class CreateCodeCLI {
     }
   }
 
-  protected async prompt() {
+  protected async _prompt() {
     const { choice } = await prompt<{ choice: string }>([
       {
         choices: ["💾  Save and Exit", "🖥️  Open REPL", "🗑️  Discard and Exit"],
```
flaurida: thought we were doing something like `🖥️  Open REPL to try out code`?

PR #390, opened 2020-02-09
Update type expression in realtime

jperl:
- [x] build VirtualCode
- [x] CodeReconciler.reconcile new expressions
- [x] use CodeReconciler in CodeFile
- [x] CodeReconciler: update existing expressions'
- [x] include selectors in repl context


packages/build-code/src/StepExpression.ts
```diff
@@ -0,0 +1,78 @@
...
+  public constructor(step: Step, previous?: StepExpression) {
+    this._previous = previous;
+    this._step = step;
+
+    if (!SUPPORTED_ACTIONS.includes(step.action)) {
+      throw new Error(`StepExpression: invalid action ${step.action}`);
+    }
+  }
+
+  private _buildOptions(): string | undefined {
```
flaurida: if it can return nothing should be `string | void`

packages/build-code/src/StepExpression.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // 1) we need to specify the page if the step has a different page
+    // 2) we don't need to, but we specify the page if it is not 0
+    //    for clarity and in case they delete steps
+    if (this._didPageChange() || page !== 0) {
+      return `{ page: ${page} }`;
+    }
+  }
+
+  private _buildValue(): string | undefined {
```
flaurida: same here

PR #402, opened 2020-02-12
✏️Create in an existing file

### LINKED ISSUE #376: ✏️Create in an existing file

jperl:
Resolves #376

To Do

- [ ] `await create()`
- [ ] `browser.recordEvents()`


jperl:
Will replace with a new PR soon

PR #408, opened 2020-02-13
Hoist dev dependencies

jperl:
- All packages use the same version of a given dependency
- Dependency installation time is reduced
- Less storage is needed


jperl:
Does not seem to work with typescript. Going to wait for yarn 2 / yarn to be figured out and maybe use workspaces or lerna and tsc improvements.

Issue #410, opened 2020-02-14
🌎Specifying browsers fails on Windows


lllauri:
**Context**
Windows 10, version 1909
Tried Windows Terminal, Powershell, cmd

**Issue**

> **PS > npx qawolf test --firefox**
> QAW_BROWSER=firefox npx jest --config={} --preset="@qawolf/jest-plugin" --rootDir=.qawolf --testTimeout=60000
> 
> 'QAW_BROWSER' is not recognized as an internal or external command,
> operable program or batch file.
> **PS > npx qawolf test --all-browsers**
> QAW_BROWSER=chromium npx jest --config={} --preset="@qawolf/jest-plugin" --rootDir=.qawolf --testTimeout=60000
> 
> 'QAW_BROWSER' is not recognized as an internal or external command,
> operable program or batch file.
> **PS > npx qawolf test**
> npx jest --config={} --preset="@qawolf/jest-plugin" --rootDir=.qawolf --testTimeout=60000
> 
>  PASS  .qawolf/tests/x.test.js (7.225s)
>  ...
> Time:        12.539s
> Ran all test suites.

jperl:
Thank you for this issue. I will fix this shortly.

jperl:
@lllauri A fix for this was merged in #414 and will be in the next version of QA Wolf. I will let you know when we release the next version, which should be later this week.

jperl:
@lllauri This is now fixed in 0.9.3. Please let me know if you have further issues.

PR #414, opened 2020-02-14
fix: 🌎Specifying browsers fails on Windows #410

jperl:
- [x] Upgrade to latest playwright
- [x] Add separate non-browser test workflow. Separate "test:browser" and "test"
- [x] Fix windows tests


jperl:
We will set up e2e tests as part of #419

Issue #419, opened 2020-02-16
✅Add comprehensive tests against sandbox (Material UI, etc)


flaurida:
Make sure issues raised in https://github.com/ZlydeTJ/qawolf-example-tests are resolved

Relates to #412

- [x] e2e
- [ ] click
- [x] type
- [ ] select
- [ ] scroll

Issue #416, opened 2020-02-16
⌨️ Do not clear in type by default ⌨️ "type(..., { replace: true})" when entire text value is selected


flaurida:
"I usually change the values a bit. If I append the filled textfield it replaces the values." from @ZlydeTJ #412

- Do not clear in `type`, remove `options.skipClear`
- Add `type` `options.replace` which selects all before typing. Therefore replacing the text.
- In create, automatically set `options.replace=true` if user clicks to select all, or cmd+a (mac), or ctrl+a (linux/windows) to select all

jperl:
`isClearSequence`: a sequence of type events, that result in the element being cleared (whereas it was not before hand). these events should be ignored in the build type steps.

This is true for: multiple backspaces, select all and delete, cmd+a delete.

Still need to figure out "select all and type".

jperl:
We could instead do something like `selectRange(selector)` which would be more flexible. However the most common use case is just to clear something. I would prefer my code to be `clear(selector); type(selector, valueWithoutBackspace)`. So that is why I think we should stick with `clear`.

jperl:
We could also do `browser.type(selector, { selectAll: true })` if CMD+A or selection range is all.

jperl:
` { replace: true }` sounds better and is still accurate.\

jperl:
Make sure we have tests for content editable.

jperl:
Make sure to test date inputs

```
Evaluation failed: DOMException: Failed to execute 'setSelectionRange' on 'HTMLInputElement': The input element's type ('date') does not support selection.
```

jperl:
Add test typing for date inputs. There is an issue with sendCharacters vs type, we should find out why.

PR #421, opened 2020-02-17
🏖️ Add sandbox with material-ui and other common frameworks for tests

### LINKED ISSUE #415: 🌲 Use attribute on ancestor as selector
Create a css selector with the data attribute targeting the descendant target.

Relates to #412

flaurida:
Sandbox with various UI elements, will be used to test #415, #416, etc.

Relates to #419


jperl:
You should revert the code changes, and submit the sandbox as its own PR.

jperl:
Also this is going to be awesome, now we can replace the-internet tests with this.

sandbox/public/index.html
```diff
@@ -40,4 +47,15 @@
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
   </body>
+  <!-- Include the Quill library -->
+  <script src="https://cdn.quilljs.com/1.3.6/quill.js"></script>
+
+  <!-- Initialize Quill editor -->
+  <script>
+    if (window.location.pathname === "/content-editables") {
```
jperl: This won't work when you first navigate to the homepage. Is there a way to do this in the component?

flaurida: Yeah I know...let me research it.

jperl:
Instead of using quill, should we just do a native content editable?

sandbox/public/index.html
jperl: I think we should nix quill, and just add a native one. From my brief using it, its pretty close to native.

jperl: Originally, I would suggest just nixing Quill and adding a native one. However since it has the UI controls it may be nice to have.

PR #426, opened 2020-02-18
⌨️ Do not clear in type by default 🏖️ Add comprehensive tests for type against sandbox

### LINKED ISSUE #416: ⌨️ Do not clear in type by default ⌨️ "type(..., { replace: true})" when entire text value is selected

jperl:
**BREAKING CHANGE** browser.type no longer clears the element by default

removed: type options.skipClear since it is no longer needed

added: type options.replace to select the element's content first. It will then type and replace the value. This allows us to accurately represent CMD+A + typing which never clears an element.

To Do

- [x] Add selectElementContent and test against sandbox
- [x] Rewrite type tests for selectElementContent against sandbox

Resolves part of #416.

I will create a separate PR that automatically creates `options.replace = true` for CMD+A/mouse selecting.


packages/browser/src/actions/typeElement.ts
```diff
@@ -26,10 +19,20 @@ export const typeElement = async (
 
   await elementHandle.focus();
 
-  const strokes = deserializeStrokes(value || "");
+  if (options.replace) {
+    await selectElementContent(elementHandle);
+  }
+
+  // default value to backspace to clear if null or "" is passed
+  const text = value || "↓Backspace↑Backspace";
```
flaurida: should only happen if `replace: true` to be consistent

jperl: Great point!

PR #428, opened 2020-02-19
⌨️ type(..., { replace: true}) when entire text value is selected

### LINKED ISSUE #416: ⌨️ Do not clear in type by default ⌨️ "type(..., { replace: true})" when entire text value is selected

jperl:
Inline `type(..., { replace: true })` when all the text is selected first

- [x] Port recorder tests to sandbox
- [x] Record selectall for text input
- [x] Remove select all key events
- [x] Add options.replace to buildTypeSteps when selectall precedes typing
- [x] Inline `{ replace: true}` for type steps

Relates to #426
Resolves #416


jperl:
~~Record selectall for content editables~~ Punting on this for now since it is a corner case and not worth the complexity yet. We would need to [combine two events](https://stackoverflow.com/a/13286950/230462). When someone brings it up we can revisit.

packages/build-code/src/StepExpression.ts
```diff
@@ -20,12 +20,32 @@ export class StepExpression implements Expression {
   private _buildOptions(): string | void {
     const page = this._step.page;
 
+    let options: any = {};
```
flaurida: type?

packages/build-workflow/src/removeShortcutKeyEvents.ts
```diff
@@ -0,0 +1,82 @@
...
+export const removeShortcutKeyEvents = (
+  eventName: string,
+  events: ElementEvent[]
+) => {
+  const filteredEvents = events.slice(0);
+
+  let key: string;
+  if (eventName === "paste") {
+    key = "v";
+  } else if (eventName === "selectall") {
```
flaurida: cut? ✂️

packages/build-code/src/StepExpression.ts
jperl: fixed 👍

packages/build-workflow/src/removeShortcutKeyEvents.ts
jperl: What do you mean cut, it's not cut. It's selectall.

jperl: We do not currently create code for cut, nor do I think it makes much sense to have cut in a test.

PR #433, opened 2020-02-21
♻️Add browser argument and export script method

### LINKED ISSUE #403: ♻️Add browser argument and export script method
- Add a browser argument to the generated --script method

- Move the call of the method to a main method. Create and close the browser in the main method

jperl:
Resolves #403


jperl:
This will likely get replaced when we do edit mode, but this is a good placeholder in the meantime.

PR #445, opened 2020-02-22
🧹Simplify API

### LINKED ISSUE #432: 🧹Simplify API and generate vanilla Playwright code 🍦(with a few small helpers)
Goals

- Generate vanilla Playwright code
- as small API as possible
- clear line between qawolf and playwright. do not decorate or wrap playwright helpers

Generate pure Playwright code, except in corner cases which are not possible with vanilla playwright yet (multiple pages/alternate frames).

Prerequisite: #398 being able to register our selector engine 🙏 

API:

```
const { scroll, launch, $page } = require("qawolf");

// playwright launch + capture and browser choice via env variable
const browser = await launch();

// pure playwright
const page = await browser.newPage();

// $page helper auto-waits for the page (and in the future frame)
// Can pass creation index or page object and it will auto-wait for page
const page = await $page(1 || page);

// missing methods we fill in
await scroll(page, value);
```

Implementation Ideas

- Remove wrapper code by hooking into prototype constructor of context/page. Set up global registries.

jperl:
Resolves #432

### DIFF
--- a/packages/browser/src/$page.ts
+++ b/packages/browser/src/$page.ts
@@ -1,0 +1,39 @@
+import { Page } from "playwright";
+
+const waitForPage = async (pageOrIndex?: number | Page) => {
+  if (!pageOrIndex) return waitForPage(0);
+
+  const page = pageOrIndex as Page;
+  if (page && page.click) {
+    // TODO wait for requests...
+    return page;
+  }
+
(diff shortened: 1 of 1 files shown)


jperl:
Will open new vanilla PR

PR #447, opened 2020-02-22
Improve click event recording

flaurida:
* Use both mousedown and click
* If last event is an input take that one (auto triggered input click on checkbox, etc.)


jperl:
So that the code generation is instant, I am going to remove the debounce logic. Send all events. Then handle this logic inside of buildClickSteps. The CodeReconciler will handle updating the step when the better event comes through.

jperl:
Closing this PR, will split it into two others.

PR #453, opened 2020-02-23
Refactor recorder

flaurida:


packages/types/src/event.ts
```diff
@@ -10,6 +10,7 @@ export interface ElementEvent {
 }
 
 export type ElementEventName =
+  | "click"
```
jperl: minor, do you mind abcing this list

Issue #441, opened 2020-02-22
⌨️Vanilla playwright keyboard


jperl:
- Use press for non-modifier keys
- Use fill for replace

PR #478, opened 2020-03-05
✏️qawolf.create()

jperl:


src/build-code/buildTemplate.ts
```diff
@@ -52,6 +52,7 @@ if (require.main === module) {
   (async () => {
     const browser = await qawolf.launch();
     ${buildNewContext(device)}
+    await qawolf.register(context);
```
flaurida: should we add a small comment here on what this does? just a thought

jperl: We could link to our docs?

src/create-code/CodeFileUpdater.ts
```diff
@@ -1,22 +1,54 @@
...
 
-  protected async updateCode(code: string): Promise<void> {
+  protected async _updateCode(code: string): Promise<void> {
     await outputFile(this._path, code, 'utf8');
   }
+
+  public async discard(): Promise<void> {
+    this._locked = true;
+
+    if (process.env.QAW_DISCARD === '1') {
```
flaurida: assuming default here should be to revert

src/create-code/CreateManager.ts
```diff
@@ -0,0 +1,118 @@
...
+    await Promise.all([
+      this._codeUpdater.update({ steps }),
+      this._selectorUpdater.update({ steps }),
+    ]);
+  }
+
+  public async finalize(): Promise<void> {
+    const shouldSave = await promptSaveRepl(this._codeUpdater.path());
+    if (shouldSave) {
+      await this._codeUpdater.finalize();
```
flaurida: what about `_selectorUpdater`

test/create-code/__snapshots__/CodeFileUpdater.test.ts.snap
```diff
@@ -0,0 +1,22 @@
...
+  // 🐺 CREATE CODE HERE
+  otherCode();"
+`;
+
+exports[`CodeFileUpdater update saves code with new steps 2`] = `
+"  someCode();
+  await page.click(selectors[0]);
+  await page.click(selectors[1]);
+  await page.type(selectors[2], \\"tomsmith\\");
+  await page.type(selectors[3], \\"↓Tab↑Tab\\");
```
jperl: This will change after #441

src/create-code/CreateManager.ts
jperl: There is nothing to finalize

src/create-code/CodeFileUpdater.ts
jperl: This is going to be a hidden env variable we pass from `npx qawolf create` to know to remove the file.

PR #485, opened 2020-03-06
statePath option in create cli

flaurida:


flaurida:
Will need to add `setState` from `playwright-utils` to the qawolf API

jperl:
playwright-utils@0.6.0 published, just do it in this pr

PR #487, opened 2020-03-06
Consume QAW_ATTRIBUTE

flaurida:


src/config.ts
```diff
@@ -1,3 +1,7 @@
+const DEFAULT_ATTRIBUTE =
+  'data-cy,data-e2e,data-qa,data-test,data-testid,/^qa-.*/';
```
flaurida: `^` in regex means start of expression

Issue #468, opened 2020-03-01
🍦Update documentation


jperl:
Include new guides:

- Edit mode
- Handle login
- Playwright recipes

PR #505, opened 2020-03-10
Update guides

### LINKED ISSUE #298: 🔑Login guide
Problem: Login once for all tests.

1. Create a script that logs in as your users, to run once before all the tests.

`npx qawolf create --script loginAdminUser`

Example `loginAdminUser.js`:

```
const { launch, saveContextState } = require("qawolf");

export const loginAdminUser = () => {
  const context = await launch(launchOptions);  

  // Your script can log in with the UI or set cookies/localStorage directly here

  // This will serialize and save all local storage and cookies to a temporary path
  await saveState(page, "adminUser.json");

  await browser.close();
};

(async () => {
  await loginAdminUser();
  // You can change this to login multiple users
  // await Promise.all([loginAdminUser(), loginRegularUser()]);
})();
```

2. When creating a test, reapply the state to your context.

`npx qawolf create myTest`

```
  beforeAll(async () => {
    browser = await launch();
    await setState(context, "adminUser");
  });
```

3. Before running your tests, run your login script: `node .qawolf/scripts/loginAdminUser.js && npx qawolf test`.

Will also include detailed documentation for how to do all of these things!

Relates to #376, #459

### LINKED ISSUE #405: 📋Edit guide
How to create more steps in an existing file with `await create()`

How to pause and try out commands at a step.  Should we rename this to `await pause()` or `await debug()`?

Consider how this overlaps with #298 

Depends on #376

flaurida:
Resolves #298, Resolves #405 

* relates to #468


docs/docs/add_assertions.md
```diff
@@ -7,89 +7,69 @@ In this guide we show you how to add assertions to your tests. We assume that yo
...
-it('can click "Clear completed" button', async () => {
-  await browser.click(selectors[3]);
-
-  // custom code starts
-  // get current Playwright Page instance
-  const page = await browser.page();
-  // wait until no more todo items appear
-  await page.waitFor(() => !document.querySelector(".todo-list li"));
-  // custom code ends
+await page.waitFor(() => {
```
jperl: I would stay consistent and do the single line waitFor if there is only one line

docs/docs/change_input_values.md
```diff
@@ -5,111 +5,71 @@ title: 📮 Change Input Values
...
-
-```js
-await browser.type(selectors[1], "↓Enter↑Enter");
+// change this
+await page.type(selectors['1_what_needs_to_b_input'], 'secret value');
+// to this
+await page.type(selectors['1_what_needs_to_b_input'], process.env.TODO_VALUE);
 ```
 
 ## Change input values
```
jperl: This guide seems superfluous if it does not include environment variables.

docs/docs/change_input_values.md
```diff
@@ -5,111 +5,71 @@ title: 📮 Change Input Values
...
 
 ```js
-it('can type into "What needs to be done?" input', async () => {
-  await browser.type(selectors[0], "create test!");
-});
+await page.type(selectors['1_what_needs_to_b_input'], 'create test!');
 ```
 
-However, you may want to change this input value to something else. The simplest way to do this is to change the second argument to [`browser.type`](api/browser_context/type) from the value you originally typed to your desired value:
+However, you may want to change this input value to something else. The simplest way to do this is to change the second argument to [`page.type`](https://github.com/microsoft/playwright/blob/master/docs/api.md#pagetypeselector-text-options) from the value you originally typed to your desired value:
```
jperl: This part of the guide is superfluous. We should only have the one environment variable example below, telling them they can change a value is more than obvious.

flaurida: Not sure I agree, I like being explicit about it. Someone could always skip that section if they prefer.

jperl: Can resolve this later

docs/docs/add_assertions.md
jperl: Pushed a fix

docs/docs/api/qawolf/stop_videos.mdx
```diff
@@ -0,0 +1,34 @@
...
+sidebar_label: stopVideos
+title: qawolf.stopVideos
+---
+
+import Argument from '../components/Argument';
+import Syntax from '../components/Syntax';
+
+<Syntax code="stopVideos()" />
+
+Stop recording video and wait until all video [artifacts](../../run_tests_in_ci#debug) have saved.
```
jperl: videos*... all videos*

PR #510, opened 2020-03-12
Fixes auth, By setting state before navigating

ZlydeTJ:
More an FYI PR, what went wrong.

As mentioned beforehand, I've run into an issue where I wanted to navigate to page behind a login and was presented with a login even though I set the cookies correctly. (As I could navigate there by hand).

Changing the order, when setting the state, works out. 

@flaurida I've tried to see why the unit tests fail, but I wasn't sure where to look at.


ZlydeTJ:
This was the wrong way to fix the issue. Normal apps shouldn't have this problem.

PR #511, opened 2020-03-12
Call setState before page.goto

flaurida:


jperl:
@flaurida Do we still want to do this?

flaurida:
Not at the moment, because trying to set cookies, etc. on a blank page is not valid (throws an error).

There is a separate issue as to whether we should reload the page after setting the state. I think it is probably safest to do that, but we can ticket separately.

PR #529, opened 2020-03-20
support "yarn create qawolf"

jperl:


jperl:
Seems like we could auto-detect it https://github.com/yarnpkg/yarn/pull/7127/files

PR #541, opened 2020-03-27
👀npx qawolf edit & watch mode

### LINKED ISSUE #526: ✏️npx qawolf edit filename
- Add `npx qawolf edit filename` that runs a single file.
- Unless `config.watch: false`, `npx qawolf create` and `npx qawolf edit` should re-run when the file is saved.
- On re-run: close the previous browser and repl

Based on feedback from @papoola

---

Update docs as part of this.

jperl:
Resolves #526


jperl:
After wrestling with building our own runner which works, except on windows, it is apparent we should build a jest watch plugin instead.

jperl:
Note to self

```
DEBUG=qawolf:Run* CI=true QAW_HEADLESS=false npx jest --config=ts-jest.config.json --watch --rootDir=test/.qawolf google.test.js --detectOpenHandles --testTimeout=3600000
```

postInstallMessage.js
```diff
@@ -0,0 +1,10 @@
+const { bold } = require('kleur');
```
jperl, line 1: I moved this file to js and to the top level, so we did not need to build it before using it. This prevents an issue when trying to link locally.

src/cli/editCommand.ts
```diff
@@ -0,0 +1,28 @@
+import program, { Command } from 'commander';
+import { loadConfig } from '../config';
+import { findTestPath } from '../run/findTestPath';
+import { runTests } from '../run/runTests';
+import { buildEditOptions } from '../run/buildEditOptions';
```
flaurida, line 5: abc

src/qawolf.ts
```diff
@@ -0,0 +1,22 @@
...
+  TemplateFunction,
+} from './build-code/buildTemplate';
+
+export { Config } from './config';
+
+export { create } from './create-code/create';
+
+export {
+  launch,
+  repl,
```
flaurida, line 15: abc

src/utils/Registry.ts
```diff
@@ -0,0 +1,51 @@
+import { EventEmitter } from 'events';
+import { Browser, BrowserContext } from 'playwright';
+import { watchBrowser } from '../watch/watchBrowser';
```
flaurida, line 3: abc

PR #561, opened 2020-04-08
✅Replace flaky create end-to-end test

### LINKED ISSUE #555: ✅Replace flaky create end-to-end test

jperl:
Resolves #555

### DIFF
--- a/.github/workflows/create-qawolf.yml
+++ b/.github/workflows/create-qawolf.yml
@@ -18,8 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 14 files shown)


jperl:
Tried to test watch mode but for some reason the jest in the child process does not re-run. I suspect it has to do with dangling processes?

```

  it('reruns the test on user changes', async () => {
    const code = await loadCode();

    const patchedCode = patchCode({
      code,
      patch: `console.log("patched code");`,
    });

    await writeFile(testPath, patchedCode, 'utf8');

    await waitFor(() => stdout.split('npx jest').length > 2);

    console.log(
      stdout.split('npx jest'),
      'length',
      stdout.split('npx jest').length,
    );
  });
```

PR #577, opened 2020-04-14
Support playwright-chromium, playwright-firefox, playwright-webkit

### LINKED ISSUE #574: Using playwright-chromium
Hey team. Hope all is well!

I'm curious if you plan to support using `playwright-chromium` or another sub-package instead of just the full `playwright` package. I'm seeing this warning right now, so it looks like this type of usage isn't supported.

```sh
npm WARN qawolf@0.13.1 requires a peer of playwright@^0.13.0 but none is installed. You must install peer dependencies yourself.
```

This is something that would be useful for me, and I'm curious if there are any specific limitations that are blocking this. Thanks!

flaurida:
* resolves #574

### DIFF
--- a/src/utils/launch.ts
+++ b/src/utils/launch.ts
@@ -43,13 +43,24 @@ export const parseBrowserName = (name?: string): BrowserName => {
   return 'chromium';
 };
 
-export const getBrowser = (browserName: BrowserName): BrowserType<Browser> => {
+export const getBrowserType = (
+  browserName: BrowserName,
+): BrowserType<Browser> => {
+  // We must use the browser type from the installed `playwright` or `playwright-browser` package,
+  // and not `playwright-core` since they store different browser binaries.
+  // See https://github.com/microsoft/playwright/issues/1191 for more details.
   let playwright: typeof playwrightCore;
 
   try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
     playwright = require('playwright');
   } catch (error) {
-    playwright = require(`playwright-${browserName}`);
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      playwright = require(`playwright-${browserName}`);
+    } catch (error) {
+      throw new Error('qawolf requires playwright to be installed');
+    }
   }
 
   return playwright[browserName];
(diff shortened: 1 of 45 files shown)


test/build-code/__snapshots__/buildTemplate.test.ts.snap
```diff
@@ -1,20 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`buildImports imports when typescript 1`] = `
-"import { Browser, Page } from \\"playwright\\";
+"import { Browser, Page } from \\"playwright-core\\";
```
jperl, line 4: I'm inclined to keep this as is, and let those who want to use `playwright-browser` packages with TypeScript or devices use a custom template, or they can edit the file.

test/browser.ts
```diff
@@ -1,5 +1,5 @@
 import Debug from 'debug';
-import { BrowserServer, chromium, ChromiumBrowser } from 'playwright';
+import { BrowserServer, chromium, ChromiumBrowser } from 'playwright-core';
```
jperl, line 2: You will need to separate the chromium import to playwright, since the browsers are installed at the root level packages (and our dev dependency is playwright).

test/build-code/__snapshots__/buildTemplate.test.ts.snap
flaurida: Yeah, after thinking about it agree with you. Pushed a commit to revert it.

src/utils/launch.ts
```diff
@@ -45,6 +43,18 @@ export const parseBrowserName = (name?: string): BrowserName => {
   return 'chromium';
 };
 
+export const getBrowser = (browserName: BrowserName): BrowserType<Browser> => {
+  let playwright: typeof playwrightCore;
```
jperl, line 47: Can you please leave this comment in here:

```
// We must use the browser from the installed `playwright` or `playwright-browser` package, and not `playwright-core` since they store different browser binaries. See https://github.com/microsoft/playwright/issues/1191 for more details.
```

src/utils/launch.ts
```diff
@@ -45,6 +43,18 @@ export const parseBrowserName = (name?: string): BrowserName => {
   return 'chromium';
 };
 
+export const getBrowser = (browserName: BrowserName): BrowserType<Browser> => {
```
jperl, line 46: This should be `getBrowserType` https://github.com/microsoft/playwright/blob/master/docs/api.md#browsertypelaunchoptions

jperl:
Annoying thing is that jest hides the beforeAll error https://github.com/facebook/jest/issues/8688 so you only get `Cannot read property 'close' of undefined` from the `afterAll`.

Issue #583, opened 2020-04-20
Error: Device = not available in Playwright


hollligan:
Hi guys! Thank you for a very simple testing framework!! :)
While researching the functionality I tried [this](https://docs.qawolf.com/docs/emulate_a_device).

steps:
1. npm init
2. npm init qawolf
3. (root-Gitlab CI/CD)
4. npx qawolf create --device = "iPad Mini" https://www.wikipedia.org searchTablet

result:
```
PS F:\Git_repositories\research\qawolf-device> npx qawolf create --device = "iPad Mini" https://www.wikipedia.org searchTablet
(node:80672) UnhandledPromiseRejectionWarning: Error: Device = not available in Playwright
    at Object.exports.buildImports (F:\Git_repositories\research\qawolf-device\node_modules\qawolf\build\build-code\buildTemplate.js:19:15)
    at exports.buildTemplate (F:\Git_repositories\research\qawolf-device\node_modules\qawolf\build\build-code\buildTemplate.js:57:29)
    at F:\Git_repositories\research\qawolf-device\node_modules\qawolf\build\cli\saveTemplate.js:19:24
    at Generator.next (<anonymous>)
    at fulfilled (F:\Git_repositories\research\qawolf-device\node_modules\tslib\tslib.js:110:62)
(node:80672) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:80672) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.     
PS F:\Git_repositories\research\qawolf-device>
```

flaurida:
Hi @hollligan thanks for the kind words 😊 

I just tried it, and the issue is with the extra space between `device` and `=`. So this command works:

```bash
npx qawolf create --device="iPad Mini" https://www.wikipedia.org searchTablet
```

But this command does not:

```bash
npx qawolf create --device = "iPad Mini" https://www.wikipedia.org searchTablet
```

Please let me know if removing the extra spaces fixes the issue for you!

hollligan:
I tried to delete extra spaces, but the problem remained.
[...]

PR #584, opened 2020-04-20
fix for use in PowerShell

hollligan:
#583


jperl:
Thank you for your docs PR. I believe this is covered by https://github.com/qawolf/qawolf/pull/585 however feel free to re-open if you do not think that is clear enough.

PR #612, opened 2020-05-20
fix: removed semicolon in PATCH_HANDLE

Cloudo:
it prevents me from editing tests when I have `"semi": false` in my prettier config with error
`Could not find await qawolf.create(); in caller`

### DIFF
--- a/src/create-code/patchCode.ts
+++ b/src/create-code/patchCode.ts
@@ -5,7 +5,7 @@ type PatchOptions = {
   patch: string;
 };
 
-export const PATCH_HANDLE = 'await qawolf.create();';
+export const PATCH_HANDLE = 'await qawolf.create()';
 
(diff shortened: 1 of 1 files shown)


flaurida:
Hi @Cloudo thanks for the PR!

The issue with just removing the semicolon is that it will now not be included for anyone. Our code templates use semicolons so it may be a bit odd. Let us think about it a bit more, or I am happy to go into more details on what would need to happen.

jperl:
I think we should replace `await qawolf.create()` with a comment. What do you think @flaurida?

jperl:
Hi @Cloudo I believe we fixed this issue in https://github.com/qawolf/qawolf/pull/684.

Now we replace `qawolf.create()` with a comment `// 🐺 create code here` so this should not interfere with your code formatting anymore. I will let you know when the next release is.

PR #731, opened 2020-07-12
switch html-parse-stringify back to upstream

jperl:


jperl:
Going to remove it entirely. We don't really use it anymore, was a relic from html selector times.

PR #733, opened 2020-07-12
Use qawolf.register to add qawolf script once before pages are loaded

### LINKED ISSUE #661: Update to playwright 1.2
Also
- Run tests for third party PRs

jperl:
Resolves #661 (fixes regression)

Set up the event recorder and logger at the same time, and they only emit events if a binding is found. Now we no longer rely on the order of when addInitScripts is called. And we can set up the binding once on the context, instead of per page.

Since the addInitScript is called in qawolf.register before pages are created, we no longer need to call evaluate and can remove the unnecessary initEvaluateScript helper. Calling addInitScript before pages are created prevents errors thrown when iframes close in the middle of script evaluation.


jperl:
Going to go with a different strategy. Where qawolf.register adds the script, so we do not need a followup evaluate.

PR #741, opened 2020-07-17
feat: Improve selector generation to use multiple attributes

### LINKED ISSUE #605: 🔍Use multiple CSS attributes per level
I believe this can reduce the occasional div:nth-of-type selectors as seen in https://youtu.be/2dyuveoW1ls?t=3412

Need to decide the priority of this vs ancestors and find more example test cases.

----

Try to use the best attributes of children, then parents up to the top most clickable ancestor. Do not immediately start at top most clickable ancestor.

aldeed:
Resolves #605 

## Changes

`buildSelector` will now use multiple attributes at every level to determine the probable best selector.

### Penalties
`iterateCues`, which is part of `buildSelector` logic, will now iterate in such a way as to favor better selectors at every level, before trying worse selectors at lower levels. To do this, the simple "rank" array of cue types is now a more configurable `CueTypesConfig` object, which assigns a `penalty` value to each cue type. In most cases, a selector with a low total penalty will be tried before a selector with a high total penalty, *regardless of level*.

Here's an example:
- Potential cue group 1: single target (level 0) cue with penalty 12
- Potential cue group 2: target (level 0) cue with penalty 5 + parent (level 1) cue with penalty 5

The second potential cue group will be preferred because even though the selector will have two levels, the total penalty is only 10, whereas the alternative is 12. If group 2 selects more than one element, then group 1 would be tried next.

I set the initial cue type penalties as follows. We may want to debate and adjust these. (I did put some thought into them, but they're also somewhat arbitrary.)

- alt: 20
- aria-label: 20
- class: 10
- contenteditable: 50 // High penalty because it is unlikely to be unique given that the value is always the same
- for: 5
- href: 10
- id: 5
- name: 10
- placeholder: 10
- src: 15
- tag: 40
- text: 12
- title: 20
- value: 10

### Cue iteration
`iterateCues` logic is somewhat simpler now:
1. Try single target cues with a max penalty of 0. (Currently this is only the custom attributes.)
2. Try increasingly long chains of cues, from any level, as long as the total penalty is 0. Ensure that there's always at least one cue from level 0 since otherwise the wrong element would be selected.
3. Increase max penalty by 5 and repeat the above two steps.
4. Repeat step 3 until a max, max penalty is hit, currently 100.
5. Fall back to xpath

Credit: I looked at https://github.com/antonmedv/finder code a lot while trying to find the best way to do this. I didn't end up using that code or any of its patterns exactly, but it was still inspiring.

### Selector caching
`buildSelector` now caches the selectors it builds by target. If we request a selector for a target that already has one, we make sure that the cached selector still matches it (DOM might have changed) and then return the cached one. Because `buildSelector` is called for every captured event, this should be very helpful for performance. The downside is that we'll need to keep watch to ensure the cache doesn't consume too much memory. If so, we could switch to a capped cache.

### Other enhancements
In addition to penalties, the new `CueTypesConfig` has an element list for each cue type. Many of the cue types apply only to certain elements, so rather than one-off check functions to see, for example, whether the `value` type applies, I made this more explicit.


aldeed:
@jperl At this point, it should be only the `buildSelector` tests that are failing and that's hopefully because all of the selectors that are different in the tests are now "better". But that may be subjective, so could you look through the failing tests and tell me if you think any selectors became worse? Hopefully we could fix any such problems by tweaking the `penalty` values until we like what we see, or at least that's the idea.

jperl:
At a quick glance this looks like a good approach. I will do a deep review on Sunday and try it out on a bunch of sites. Thank you 💯 .

jperl:
Going to comment as I investigate.

1) Scrolling chooses ` xpath=/html` instead of shortcutting to `html`

2) Google something, click on Images

The href (and src) is not a great attribute, since it often contains long dynamic urls. We could do something where we weight it worse if the url is not dynamic. In this case text is a better selector than href/src. I find this is often the cases for links, so we may just want to make href=12, text=11 or something like that.

src/web/cues.ts
```diff
@@ -107,85 +153,85 @@ export const buildCueValueForTag = (element: HTMLElement): string => {
...
   level,
 }: BuildCuesForElement): Cue[] => {
-  let cssAttributes = [...CSS_ATTRIBUTES];
-
-  if (!canTargetValue(element)) {
-    cssAttributes = cssAttributes.filter((attribute) => attribute !== 'value');
+  // For body and html, we never have more than one, so
+  // just 'tag' cue is needed and we can save some time.
+  const tagName = element.tagName.toLowerCase();
+  if (['html', 'body'].includes(tagName)) {
```
jperl, line 165: I think the penalty should only be 0 if it is the target (level 0).

src/web/iterateCues.ts
```diff
@@ -1,92 +1,93 @@
...
-  for (const ancestorCue of include(ancestorCues, 'attribute')) {
-    yield* iterateGroups(targetCues, ancestorCue);
+    trialGroups.push(expandedGroup);
+    trialGroupsIndexes.push([...startingGroupIndexes, index]);
   }
 
-  // iterate remaining cue types
-  for (const type of CueTypesRanked.slice(1)) {
-    yield* iterateType(targetCues, ancestorCues, type);
+  yield* trialGroups;
```
jperl, line 69: should this be sorted by penalty first?

src/web/iterateCues.ts
```diff
@@ -1,92 +1,93 @@
...
+ *   selector.
+ *
+ *   By calling this function in a loop with increasing penalty ranges,
+ *   we are able to prioritize having a low total penalty, even if that
+ *   penalty is the result of multiple cues.
+ */
+function* combineCues(allCues: Cue[], {
+  minPenalty,
+  maxPenalty,
+  startingGroup = [],
```
jperl, line 43: I like the name baseGroup better? But its not that important.

jperl:
Does this support using multiple attributes? It seems like it only supports one attribute per level still.

src/web/cues.ts
aldeed: Makes sense. What penalty do you like for non-0?

src/web/iterateCues.ts
aldeed: I believe it automatically will be because allCues is sorted by penalty and we build groups in ascending penalty order from that? Not positive about that, but I can try to confirm.

aldeed:
@jperl 

> Does this support using multiple attributes? It seems like it only supports one attribute per level still.

It should. If you look at `iterateCues.test.ts.snap`, the third group down is `['0data-qa', '0data-testid']`. Further down is `['0id', '0data-qa']`. Do you have specific examples where you expected to see two atts used? It might be that some penalties need to be adjusted so that a single isn't ranked higher than a better double.

I will fix the other issues you found, too.

And I also noticed that `iterateCues` is finding duplicate groups due to combining later cues with earlier ones, where the combination was already returned in reverse. I'll fix that, too.

src/web/cues.ts
aldeed: Changed in https://github.com/qawolf/qawolf/pull/741/commits/6450eb335cd1a4a5d9a68d405da8c647dc5ebc02. Using 0 for target wouldn't really matter since there's only one total cue for `html` target and two for `body`, both with same penalty. So for simplicity I'm just using the `tag` penalty for both.

src/web/iterateCues.ts
aldeed: Added a test to confirm that it's auto-sorted by penalty. Note that this is only true when the penalty increment is `1`, but by default I have it set to 5. Here's how I've documented this setting:

> A higher value means a greater chance that higher penalty groups are returned before lower penalty groups, but a lower value potentially does a lot more unnecessary looping. The value must be 1 or greater, and a value of 1, although slowest, ensures results are 100% sorted in total penalty order. Values above 1 cause the penalty sorting to be increasingly fuzzy.

So basically we want to set this as high as is acceptable, but it's possible that 5 is too high. Some combination of decreasing this value and adjusting the penalties is how we program this algorithm, but getting it exactly right will be difficult. Ideally we'd feed it a lot of scenarios and have it fine tune itself.

src/web/iterateCues.ts
aldeed: Changed

aldeed:
`xpath=/html` is fixed in https://github.com/qawolf/qawolf/pull/741/commits/792697316e6fb0b1aa5b77304f9beebb7ad040b5. This was an existing issue with `querySelectorAll({ parts: selectorParts }, document.body)` in `isMatch` but was previously hidden by override logic that I now removed. This was returning an empty array because `html` is *above* `document.body`. I changed `document.body` to `document` to fix it, but only after verifying in the playwright code that they're also calling it with `document` for that arg.

aldeed:
@jperl I think every issue you mentioned is fixed now, plus a couple I found.

Side thought:

I don't think we want to have a low penalty for `tag` since it's rarely unique BUT I do think that including the tag in the level 0 selector makes the code more readable. What do you think about always including the target tag in every selector, not for uniqueness but for readability? I wouldn't change the iteration and checking logic, but I'd add a level 0 selector part right before building the selector string, after we have a match, if it isn't already there?

(If we don't want this to be the default, it could be opt-in with config.)

jperl:
@aldeed

> I changed document.body to document to fix it

Makes sense, thanks for the explanation.

> Including the tag makes the code more readable

I do not think that is a good idea, since it will be worse for test maintenance. Ideally our cues have high semantic value (test attribute for example) and are robust to changes. Whereas the tag can easily change, ex change a button to input type=button, or a div -> section, the selector will no longer work.

I also don't necessarily agree it makes the selectors more readable, curious if you have an example.

> Multiple classes

I added an example to the sandbox, run the local version to see it (`cd packages/sandbox && npm i && npm start`). Click on the buttons labeled "Classes" and they will output this:

```
  await page.click(".a");
  await page.click("button:nth-of-type(7)");
  await page.click("button:nth-of-type(8)");
```

Expected if multiple classes are supported:

```
  await page.click(".a");
  await page.click(".b.a");
  await page.click(".b.b");
```

jperl:
I went through the test failures and outlined what I think we should fix:

Fix this:

```
Expected: "#nav-hamburger-menu"
Received: "xpath=//*[@id='nav-hamburger-menu']"
```

Fix this:

```
Expected: "[data-qa=\"quill\"] [contenteditable=\"true\"]"
Received: "xpath=//*[@id='root']/div[3]/div/div/div[2]/div[1]"
```

Multiple test attributes are always better than any one of something else (except one test attribute). Perhaps we should weight the penalty only once per cue type or something like that:

```
Expected: "[data-qa=\"material-date-picker\"] [aria-label=\"change date\"]"
Received: ".MuiButtonBase-root"
```

----

Tests that are ok changes:

Test attributes + target selector that now just target the unique target selector. Exs:

```
Expected: "[data-qa=\"material-text-input\"] .MuiInputBase-input"
Received: ".MuiInputBase-input"
```

```
Expected: "[data-qa=\"html-radio-group\"] #dog"
Received: "#dog"
```

Tests where classes are now prioritized over placeholder:

```
Expected: "[placeholder=\"What needs to be done?\"]"
Received: ".new-todo"
```

aldeed:
@jperl Thanks for the classes example. There are a couple things going on.

1. The main issue with your particular example is that "b" class returns `true` from the `isDynamic(c)` check so it is discarded. I do think trying to avoid dynamic classes and IDs is necessary, but maybe we need to make this check less strict? I tried classes like `test-class-1`, which were also deemed to be dynamic. Strangely, the letter "a" is not dynamic.
2. After switching the classes in your example to be all non-dynamic, it kind of works, but it does pick classes that match multiple elements. I had been thinking that `isMatch` ensures there's exactly one match, but actually it just ensures that the first match is the target. That does not seem very stable because if I then rearrange the buttons, the test breaks. (i.e., it's no better than nth-of-type). Should I add a `result.length === 1` check in `isMatch` or were you purposefully avoiding that?

I'll look into the test fixes next.

aldeed:
@jperl 

It seems `#nav-hamburger-menu` isn't used because it fails your new `isVisible` filter in `isMatch`. The hamburger menu isn't visible the way that fixture page renders.

Fixed `[data-qa="quill"] [contenteditable="true"]`.

In testing that, I discovered another issue with contenteditables (at least seemed like issue to me). When you type in the Quill box and press Enter, it makes paragraph elements within there. Then when you later click on one of these paragraphs, it does a "click" on "p" or some selector of the paragraph element. Since we "bubble up" all the value changes to the topmost contenteditable, it seemed like we should be doing the same with the "click" events. I fixed this here: https://github.com/qawolf/qawolf/pull/741/commits/5b91d1f11ab1f5647d81639dc99bf5a156de3bdf (Not sure whether this could be undesirable in some cases... contenteditables that contain html are tricky.)

And one more thought: regarding `'[data-qa="material-textarea"] .MuiInputBase-input'` vs `'.MuiInputBase-input'`, this got me thinking that maybe the custom attributes like `data-qa` should be added regardless of whether they are technically needed? If someone is adding `data-qa` attribute, I presume it is precisely for tools like this. It might be best to do this as a new opt-in config feature, a separate list of atts for which you always want the nearest one in every selector.

src/web/iterateCues.ts
jperl: Ok but even if you set the increment to 5 and if this had a sort right here wouldn't it still be sorted? In that case, I think we should just do a sort right here.

```
trialGroups.sort((a, b) => a.penalty - b.penalty);
yield* trialGroups;
```

jperl:
> The main issue with your particular example is that "b" class returns `true` from the `isDynamic(c)` check so it is discarded.

Ooops. I will look into making sure `test-class-1` and the other one is not marked as dynamic.

> Should I add a `result.length === 1` check in `isMatch` or were you purposefully avoiding that?

This is a tough one. I purposefully avoided it because I have run into multiple times in the real world where there are not other selectors that will work. The other option is to make multiple matches a penalty?

> It seemed like we should be doing the same with the "click" events. I fixed this here: 5b91d1f

That looks good. Please just make sure to add a test 🙏.

> Custom attributes like data-qa should be added regardless of whether they are technically needed

I agree with this.

----

Feels like this is close to the finish line. Thank you for all your great work and patience on the review 🙏, was taking a long overdue vacation last week.

src/web/cues.ts
jperl: Thanks. Marking as resolved.

jperl:
> The main issue with your particular example is that "b" class returns `true` from the `isDynamic(c)` check so it is discarded.

I added `test-class-1` to the isDynamic tests and it is not marked as dynamic. I updated the sandbox to use non-dynamic classes in 05be1de. However it still generated this:

```
  await page.click(".btn-a");
  await page.click(".btn-b");
  await page.click("button:nth-of-type(8)");
```

aldeed:
@jperl See the latest batch of commits. I resolved test issues, added missing tests, and updated it to always include the nearest preferred attribute if there is one.

I'm still seeing a few strange test failures locally, timeout stuff, but I don't have time to look into it right now.

Plus there is still the failing test for `#nav-hamburger-menu` due to the `isVisible` check now. Should we just delete that test?

aldeed:
Oh, and I do think we should add some sort of penalty to `isMatch` multiple. Maybe just tell it to keep trying until it hits N groups that work (10?), then take the one with the fewest matches, but stop as soon as there's one with exactly 1 match. But I will create a separate issue for that.

jperl:
Yes go ahead and delete that test 👍.

Any thoughts on this comment? https://github.com/qawolf/qawolf/pull/741#discussion_r464088149

Also I am still running into this issue https://github.com/qawolf/qawolf/pull/741#issuecomment-667696905

src/web/iterateCues.ts
aldeed: @jperl With increment > 5, still only that batch would be sorted (which it already is). This is because `combineCues` is recursively adding specificity, but it yields between each recursive call.

An example where sorting would be incorrect is this:

Penalty range: 1-5

1. First pass results in cue group A with total penalty 0, and cue group B with total penalty 5.
2. Only group B is in trialGroups because group A is outside the range
3. Group B isn't a match so we call recursively with each base group
4. Second pass, group A now has total penalty 1 and group B has 6.
5. Only group A is in trialGroups, but its total penalty is 1, which is less than the previous trial group penalty of 5.

The net effect of the outer loop is:
1. Try this group with one cue but higher total penalty
2. Now try this group with two cues but lower total penalty.

We could do all the recursion before any yielding, and then sort before yielding, but that will negate some of the benefits of generator function. If the element is very deeply nested, I feel like this would be very slow.

We could increment penalty ranges by 1, but then we loop many more times through the cue list. I could performance test to see which penalty increment provides the best speed vs. sort balance, but it depends a lot on the specific page/DOM, so we'd need to run it across hundreds of representative pages/actions.

aldeed:
@jperl Deleted that test and replied to that comment.

Regarding:

```
await page.click(".btn-a");
await page.click(".btn-b");
await page.click("button:nth-of-type(8)");
```

`".btn-b"` is used because it does match the `btn-b btn-a` button, even though there are multiple results (this is why I asked whether it should check for `length === 1`). This would be fixed by the solution to the separate issue I filed.

`"button:nth-of-type(8)"` is used because `btn-b btn-b` is invalid. When there are two identical classes, browsers dedupe them and make it look as though there is only one. Therefore the only possible class selector is `.btn-b`, but that already matches the button above, so it falls back to `nth` instead.

src/web/iterateCues.ts
jperl: Ok seems like we should just not worry about optimizing this, until we run across more examples in the wild

jperl:
Wow I completely overlooked that, sorry that was a bad example. Just updated it and confirmed it works as expected -- using multiple classes to disambiguate.

```
  await page.click(".btn-a");
  await page.click(".btn-2");
  await page.click(".btn-b");
  await page.click(".btn-b.btn-2");
```

PR looks great 🚀. Merge when you are ready and I will update to playwright@1.3.0 and cut a new qawolf release.

PR #766, opened 2020-08-10
Update dependencies (including playwright 1.3.0). Depend on playwright instead of playwright-core

flaurida:
This change also makes playwright a dependency since we are having issues with playwright-core types mismatching with playwright types. We ran into [issues](https://github.com/microsoft/playwright/issues/2540) in playwright@1.2.0 and it seems there is limited support for playwright-core, so dropping support for the playwright-browser flavor packages for now. At least until upstream playwright-core is stable.


test/browser.ts
```diff
@@ -1,5 +1,5 @@
 import Debug from 'debug';
-import { chromium } from 'playwright';
+import { chromium } from 'playwright-core';
```
jperl, line 2: we should be importing the browser from 'playwright'

test/web/attribute.test.ts
```diff
@@ -1,4 +1,4 @@
-import { Browser, Page } from 'playwright';
+import { Browser, Page } from 'playwright-core';
```
jperl, line 1: prefer to import from playwright to make sure the types are compatible

PR #773, opened 2020-08-11
fix: cue order

jperl:


src/web/cues.ts
```diff
@@ -243,7 +283,9 @@ export const buildCues = ({ attribute, isClick, target }: BuildCues): Cue[] => {
 
 export const buildSelectorParts = (cues: Cue[]): SelectorPart[] => {
   const levels = [...new Set(cues.map((cue) => cue.level))];
-  levels.sort().reverse();
+
```
jperl, line 286: this is the only non-stylistic change

PR #783, opened 2020-08-16
fix: id and class cues with special characters

jperl:


src/web/cues.ts
```diff
@@ -194,15 +194,25 @@ export const buildCuesForElement = ({
         element.classList.forEach((c) => {
           if (isDynamic(c)) return;
 
-          list.push({ level, penalty, type: 'class', value: `.${c}` });
+          list.push({
```
flaurida, line 197: are these special characters only possible on `id` and `class` attributes?

jperl: From my testing attribute special characters like in placeholders, do not need escaping. When I do escape, it adds unnecessary characters (which still work, but looks worse). If we run into a future bug though we can escape them.

PR #784, opened 2020-08-16
IFrame support

### LINKED ISSUE #279: 🖼️Support iframes
Name page variables as part of this, since we will need to track variables.

aldeed:
Resolves #279 

This adds support for events that occur in iframes, and also adjusts the page variable handling so that each page gets its own variable (as does each iframe), rather than overwriting `page` variable. Page and frame variables are initialized only once, on the line immediately before the first line that needs them.

The sandbox has a new `/iframes` route for manual testing, and used by some new Jest tests. Here's an example test generated from that route:

```js
test("localhost", async () => {
  await page.goto("http://localhost:5000/iframes");
  await page.click("#root div");
  await page.click('text="Main Page Button"');
  const frame = await (await page.$('[data-qa="first"]')).contentFrame();
  await frame.click('[data-qa="html-button-with-children"]');
  await qawolf.scroll(frame, "html", { x: 0, y: 100 });
  const frame2 = await (await page.$('[data-qa="second"]')).contentFrame();
  await frame2.click('[data-qa="html-text-input"]');
  await frame2.fill('[data-qa="html-text-input"]', "foo");
  await frame2.press('[data-qa="html-text-input"]', "Tab");
  await frame2.fill('[data-qa="html-text-input-filled"]', "bar");
  await frame2.press('[data-qa="html-text-input-filled"]', "Tab");
  await qawolf.scroll(frame2, "html", { x: 0, y: 158 });
  await page.click("#root div");
  await page.click('text="Main Page Button"');
  await frame.click(".second-half.type-two");
  await frame.click(".second-half");
  await frame.click("#submit-input");
});
```

Notes:
- I did not test nested iframes more than one level deep. That is out of scope.
- ⚠️ This is subject to cross-origin limitations. If the iframe loads a source on a different domain, it doesn't work, and I don't believe there is any way for it to work. For example, filling out the example inputs on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input does not work. The input examples are iframes, but they are loaded from a different domain.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1248,9 +1248,9 @@
       }
     },
     "@qawolf/sandbox": {
-      "version": "0.1.18",
-      "resolved": "https://registry.npmjs.org/@qawolf/sandbox/-/sandbox-0.1.18.tgz",
-      "integrity": "sha512-hIk0WCpGWCafhq/WYXYpgnc/3hJVp+fyfuXan109Lz1xtebiyEVLfog+7lj+7cCUZxZ+Z1Q0g5ze2NKNC9cLJA==",
+      "version": "0.1.19",
+      "resolved": "https://registry.npmjs.org/@qawolf/sandbox/-/sandbox-0.1.19.tgz",
+      "integrity": "sha512-Plg0VMJlIuOL9H/HU4Oyo1v3O8qDyZ5lwJUCuD060sY76BpG/+BHDJm8JunbaSiIIZOSDGYDIqIcWdMRAN8CEQ==",
       "dev": true,
       "requires": {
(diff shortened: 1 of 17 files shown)


aldeed:
@jperl Any idea why the tests are failing?

jperl:
> @jperl Any idea why the tests are failing?

The sandbox needs to be published. I'll do that shortly.

jperl:
@aldeed https://github.com/qawolf/qawolf/pull/784/commits/6ccdd40f5c3a98662642a874658c1963732fb7a2 should fix xdomain support (for non-nested iframes)

PR #803, opened 2020-08-28
fix: fill correct value for contenteditable inputs

### LINKED ISSUE #801: Broken for inputs that do not emit value until blur
On Behance when you search the input does not emit a change event until it blurs (you take another action). After taking the next action, like click, it will create another line of code and not update the search fill value.

aldeed:
- Create `getInputElementValue` function with adjusted logic
- Create tests for it

Fixes #801. Refer to the problem description [here](https://github.com/qawolf/qawolf/issues/801#issuecomment-683141690)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1248,9 +1248,9 @@
       }
     },
     "@qawolf/sandbox": {
-      "version": "0.1.20",
-      "resolved": "https://registry.npmjs.org/@qawolf/sandbox/-/sandbox-0.1.20.tgz",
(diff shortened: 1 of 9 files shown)


aldeed:
@jperl One of my new tests is failing on CI checks with "element not found". It passes locally. This would probably mean it's running against an older sandbox on CI (because it relies on the element I added to sandbox). But I see a sandbox build happening before the tests run, so let me know if there's something I'm no doing correctly here.

jperl:
@aldeed I will fix that and merge. Thank you!

PR #805, opened 2020-08-29
feat: support for dblclick

### LINKED ISSUE #802: Record multiple clicks (double click as one action)

aldeed:
Resolves #802 

## Changes
When a browser registers a 'dblclick', generated Playwright code will reflect this by calling `.dblclick` instead of two lines of `.click`.

## Testing
1. Double-click an element and verify that the generated code has one `dblclick` line rather than two `click` lines.
2. Click an element twice, slowly, and verify that the generated code has two `click` lines.

### DIFF
--- a/src/build-code/VirtualCode.ts
+++ b/src/build-code/VirtualCode.ts
@@ -32,8 +32,10 @@ export class VirtualCode {
           compareToLine,
           existingLineIndex
         );
+        // remove all existing lines after the first changed line
         removedLines = this._lines.slice(existingLineIndex);
-        debug('found removed lines: %j', removedLines);
+        debug('remove lines: %j', removedLines);
+        // and on the compareTo side, we now know that every line after this is new
         newLines = compareToLines.slice(existingLineIndex);
         break;
       }
@@ -41,6 +43,8 @@ export class VirtualCode {
 
     if (removedLines.length === 0) {
       debug('no changed lines');
+      // if compareToLines started with all the same lines as we previously
+      // had, then every additional line (if any) is new.
       newLines = compareToLines.slice(this._lines.length);
     }
 
@@ -49,7 +53,7 @@ export class VirtualCode {
       return null;
     }
 
-    debug('found new lines: %j', newLines);
+    debug('new lines: %j', newLines);
     return { newLines, removedLines };
   }
 
--- a/src/build-code/VirtualCode.ts
+++ b/src/build-code/VirtualCode.ts
@@ -3,8 +3,8 @@ import Debug from 'debug';
 const debug = Debug('qawolf:VirtualCode');
 
 type LinePatch = {
-  original: string;
-  updated: string;
+  newLines: string[];
+  removedLines: string[];
 };
 
 export class VirtualCode {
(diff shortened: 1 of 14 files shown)


aldeed:
@jperl This isn't working fully yet and needs more tests. But I could use your advice on how to do the rest.

So far the dblclick is recognized and `buildClickSteps` adds the `dblclick` action. However, since the first `click` goes all the way through and is added to the test code first, then after-the-fact we know it was actually a double click, I ended up having to overwrite the previous click step.

This works to a point, but the generated test isn't actually correct because in `buildPatch` this causes us to hit "last line no longer exists, wait to update".

So generally speaking, how do you think we should handle a case like this where we essentially need to "erase" an already-added line? The other option is to debounce somewhere for a second or two before generating click lines, so we can verify it wasn't the first click of a double. We could even debounce before `sendEvent` on the `web` side, but I don't know if this might make the code generation appear sluggish, or cause out-of-order events, or some other downside.

jperl:
The reconciler should be able to update the last line, that is how fill works. I would be curious why it does not work in this case.

aldeed:
@jperl It appears to be because in addition to the last line changing, the next-to-last line is now gone.

```js
 this._lines [
      'await page.click("text=Large");',
      `await page.click('text="Item 4"');`,
      `await page.click('text="Item 4"');`
    ]
```

```js
compareToLines [
      'await page.click("text=Large");',
      `await page.dblclick('text="Item 4"');`
    ]
```

After we've determined it was a double click, we essentially merged the last two "click" lines into one "dblclick" line.

This is the check that is hit:

```js
if (lastIndex >= compareToLines.length) {
      // if the last line no longer exists
      // we will update it when a new line arrives
      debug('last line no longer exists, wait to update');
      return null;
    }
```

I can get it to work if I remove that code block and adjust a few other things, but I am not sure what other issues that would cause.

aldeed:
I've reworked the `CodeReconciler` and `patchCode` to be able to patch more than just the last line. It looks for the first line that differs when comparing, and then inserts all new lines from there. Pretty sure this won't cause any other issues. I'll finish it up and push tomorrow.

jperl:
@aldeed That will work if you update the last lines moving backwards. Not forwards, since people may add assertions to the code.

aldeed:
@jperl I believe I've implemented it in a way that supports existing code having been altered (at least as much as it was supported before this). I pushed it along with new and updated tests, so this is ready for review.

This should work for 99.9% of cases, but if they've altered the selector or something on the last line or added a new last line, we might have problems. This seemed to be true even before my changes. Someday we might want to think about just generating the whole new file or test block, running both files through a git merge, and if there are conflicts, just save the file with conflict markers for them to resolve manually.

src/build-code/VirtualCode.ts
```diff
@@ -15,37 +15,42 @@ export class VirtualCode {
...
+      const existingLine = this._lines[existingLineIndex];
+      const compareToLine = compareToLines[existingLineIndex];
+      if (existingLine !== compareToLine) {
+        debug(
+          'first changed line went from "%s" to "%s" at index %d',
+          existingLine,
+          compareToLine,
+          existingLineIndex
+        );
+        removedLines = this._lines.slice(existingLineIndex);
```
jperl, line 35: Maybe a comment "// remove all lines after the first changed line" would be helpful

src/build-code/VirtualCode.ts
```diff
@@ -15,37 +15,42 @@ export class VirtualCode {
...
+        debug('found removed lines: %j', removedLines);
```
jperl, line 36: It did not really find all the lines to remove, so maybe this should just be `'remove lines:  %j`

src/build-code/VirtualCode.ts
```diff
@@ -15,37 +15,42 @@ export class VirtualCode {
...
+    }
+
+    if (newLines.length === 0) {
+      debug('no new lines');
       return null;
     }
 
-    debug('last line changed from "%j" to "%j"', lastLine, compareToLastLine);
-    return { original: lastLine, updated: compareToLastLine };
+    debug('found new lines: %j', newLines);
```
jperl, line 52: for consistency might as well just be `'new lines: %k'

src/build-workflow/buildClickSteps.ts
```diff
@@ -65,13 +65,37 @@ export const buildClickSteps = (events: ElementEvent[]): Step[] => {
...
-      event = inputEvent;
+    // Use the first "click" event in the group.
+    // Except if an event in the group is on an input, assume the click propagated
+    // to an element like a checkbox or radio, which is most accurate target.
+    const event = firstEventWithInputTarget || firstClickEvent || events[0];
+
+    // If there's a click that is the second in a double-click series,
+    // replace the already-pushed first click with a dblclick.
+    if (firstDoubleClickEvent && steps.length) {
+      steps[steps.length - 1] = {
```
jperl, line 93: How do we know this is replacing the click instead of the mousedown. What if there are two click events and then the double click.

Maybe it should filter out any clicks then append the double click instead?

jperl:
Left a few minor comments, besides those the `ContextEventCollector` should have another test that it collects double click (might need to update sandbox).

jperl:
> This should work for 99.9% of cases

Yes this is good enough probably forever. Git merge conflict is probably overkill, especially compared with the other improvements we can make to the experience.

> Re the PR

Left a few minor comments, besides those the `ContextEventCollector` should have another test that it collects double click (probably need to update sandbox with one) then 👍.

src/build-workflow/buildClickSteps.ts
aldeed: @jperl At this point everything has already been grouped, so is it possible for there to be any mousedowns? Maybe only if they click down but never let go?

Also I don't think it matters how many clicks came before because a click with `isDoubleClick===true` will _always_ immediately follow a normal click. `isDoubleClick` is a marker that means "event was 'click' and click count was 2". Maybe this is relying too much on assumptions from outside the scope of this function, though?

In practice, it seemed to always work, but I'll try breaking it with more complicated clicking frenzies. 😄 And if you can list out all the event sequences that you're worried about, maybe I can add a test for each?

aldeed:
Resolved code comment and debug message issues. See response to other comment, and I'll add `ContextEventCollector` test later today.

src/build-workflow/buildClickSteps.ts
jperl: Sorry I was tired when I reviewed this. I thought this had to do with grouping. Your logic is correct.

PR #826, opened 2020-09-06
Fix click selector penalties

### LINKED ISSUE #795: Crawling logic should prioritize selector penalty
[`getClickableAncestor`](https://github.com/qawolf/qawolf/blob/d4c49cf138842bded2a7c534430963b4dde7fcde/src/web/PageEventCollector.ts#L94-L97) is called before the selector is built. This will crawl up to the top most clickable ancestor, and then build the selector for that target. This is not good, because the lowest penalty selector might be on the target or one of the clickable ancestors that is not the top most. I have run into multiple examples where it skipped past a good potential selector ("div:nth-of-type(2) div div div" instead of "text=Skip"). 

Instead of choosing the top most element to select, we should pick the selector with the lowest penalty from any of {target,...,top most clickable ancestor}. The way I would implement this is to pass the target and any eligible targets to buildSelector. [`trimExcessCues`](https://github.com/qawolf/qawolf/blob/d4c49cf138842bded2a7c534430963b4dde7fcde/src/web/optimizeCues.ts#L247) on the target then update [`findBestCueGroup`](https://github.com/qawolf/qawolf/blob/d4c49cf138842bded2a7c534430963b4dde7fcde/src/web/optimizeCues.ts#L253) to  accept multiple targets and allow it to [`match`](https://github.com/qawolf/qawolf/blob/d4c49cf138842bded2a7c534430963b4dde7fcde/src/web/optimizeCues.ts#L221) any of them.

aldeed:
Resolves #795 

## Changes

Previously, for "click" events, the generated selector would target the topmost element in the clickable group. For example, if "click" is received for `button > span > i` element, the selector would be generated for that `button`.

However, sometimes the original target or one of the in-between elements has a better attribute for selection. For example, maybe one of them has a `data-qa` attribute.

Now, all elements in the clickable group are considered when generating a selector, and the one with the lowest penalty will be used.

## Testing
1. Run sandbox website in this branch.
2. `DEBUG=qawolf:* qawolf create http://localhost:5000/buttons`
3. Click the "Better attribute on span" button
4. Verify that `await page.click('text="Better attribute on span"');` line is added to the generated test file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1248,9 +1248,9 @@
       }
     },
     "@qawolf/sandbox": {
-      "version": "0.1.21",
-      "resolved": "https://registry.npmjs.org/@qawolf/sandbox/-/sandbox-0.1.21.tgz",
-      "integrity": "sha512-4uLawdQeTEWLM7kGzq45OJ3/oqsxMy2FoN4fyNllnl41e95hGjeD99xZobkdNmOTyeaYaHFJmB1e3Z5EYGUhAQ==",
+      "version": "0.1.22",
+      "resolved": "https://registry.npmjs.org/@qawolf/sandbox/-/sandbox-0.1.22.tgz",
+      "integrity": "sha512-SAztrrrA/d3J27ug87N/KDoQpPMAnw8OvKdHyj6hW5IVSM3KOTi+OCi3V9LiDXVQl9iXem30Qs3aG+TS2UTNtA==",
       "dev": true,
       "requires": {
         "serve": "^11.3.0"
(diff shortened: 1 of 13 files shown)


aldeed:
@jperl This is ready for review. My only small concern is that maybe there are some edge cases where it would be beneficial to slightly penalize the non-topmost elements in the clickable group. Basically we are losing the "prefer topmost" aspect because all elements in the group are now equally preferred, prior to applying the selector penalties.

But that being said, if this fixes all the issues you've seen, the we can always improve it later.

jperl (approved):

PR #832, opened 2020-09-08
fix: handle more nested button group cases

### LINKED ISSUE #827: Improve click group
Right now we crawl up to include all clickable ancestors, we should consider descendants as well -- stopping at the same boundaries (`a`, `button`, `input`) ideally w/ bubbling.

aldeed:
Resolves #827 

## Changes

When finding the nicest selector for a click event:
- Other descendants of the topmost clickable element will be considered, even if they are not ancestors of the original click `target`. This currently goes to a maximum depth of 10 elements down from the topmost element
- `svg` elements will be considered but none of their descendants are considered because there could be many
- There is a boundary between clickable groups nested within other clickable groups. For example, if there is an `<a>` within a `<button>` and you click on the `<button>` outside of the `<a>`, neither the `<a>` nor any of its descendants are considered as possible selectors for the click event. Conversely, if you click on the `<button>` within the `<a>`, then the `<a>` and all its descendants are considered, but the `<button>` ancestor and its group elements are not considered.
- An element is deemed to be the topmost element of a clickable group if ANY of the following are true (see `isLikelyTopOfClickGroup` function):
    - is `a`, `button`, `input[type=button]`, `input[type=image]`, `input[type=reset]`, `input[type=submit]`
    - has `role="button"` attribute
    - has no parent
    - has a non-clickable parent

Also optimized the final picking (sorting) of the best cue group in a new function, `pickBestCueGroup`, not to be confused with the existing `findBestCueGroup`.

### DIFF
--- a/src/web/element.ts
+++ b/src/web/element.ts
@@ -72,14 +72,15 @@ const traverseClickableElements = (
   if (direction === 'down' && depth > 0 && isLikelyTopOfClickGroup(element)) return;
 
   const newDepth = depth + 1;
-  if (newDepth > maxDepth) return;
-
   const lowerTagName = element.tagName.toLowerCase();
 
   if (direction === 'up') {
     // Call self for the parent element, incrementing depth
     traverseClickableElements(element.parentElement, group, direction, maxDepth, newDepth, [lowerTagName, ...ancestorChain]);
   } else {
+    // Respect max depth only when going down
+    if (newDepth > maxDepth) return;
+
     // If we make it this far, this element should be part of the current group.
     // We add elements to the group only on the way down to avoid adding any twice.
     group.push(element);
--- a/src/web/element.ts
+++ b/src/web/element.ts
@@ -1,5 +1,8 @@
 import { getXpath } from './serialize';
 
+const BUTTON_INPUT_TYPES = ['button', 'image', 'reset', 'submit'];
+const MAX_CLICKABLE_ELEMENT_TRAVERSE_DEPTH = 10;
+
 export const isVisible = (
   element: Element,
   computedStyle?: CSSStyleDeclaration,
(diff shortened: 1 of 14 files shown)


src/web/element.ts
```diff
@@ -22,14 +25,79 @@ export const isVisible = (
 
 export const isClickable = (
   element: HTMLElement,
-  computedStyle: CSSStyleDeclaration,
+  computedStyle: CSSStyleDeclaration = window.getComputedStyle(element),
```
jperl, line 28: I didn't know default values could reference variables

src/web/element.ts
```diff
@@ -22,14 +25,79 @@ export const isVisible = (
...
+  maxDepth: number = MAX_CLICKABLE_ELEMENT_TRAVERSE_DEPTH,
+  depth = 0,
+  ancestorChain: string[] = [],
+): void => {
+  // Regardless of which direction we're moving, stop if we hit a non-clickable element
+  if (!isClickable(element)) return;
+
+  // When moving up, when we reach the topmost clickable element, we
+  // stop traversing up and begin traversing down from there.
+  if (direction === 'up' && isLikelyTopOfClickGroup(element)) {
```
jperl, line 66: I think this will build an empty group if it traverses up to the max, since it will never go back down. Perhaps it should be `direction === 'up' && (isLikelyTopOfClickGroup(element) || depth === maxDepth)`

src/web/optimizeCues.ts
```diff
@@ -261,20 +251,28 @@ export const optimizeCues = (
...
-      if (a.valueLength > b.valueLength) return 1;
+export const pickBestCueGroup = (cueGroups: CueGroup[]): CueGroup | null => {
+  let bestCueGroup: CueGroup;
 
-      return 0;
-    });
+  if (cueGroups.length === 0) return null;
+
+  // Rank by the total penalty then by value length. This will take less
+  // time than .sort and will not mutate the cueGroups array.
```
jperl, line 266: 👍

src/web/element.ts
aldeed: 👍 I believe they can only reference previous arguments, but that's from memory so I might be wrong.

src/web/element.ts
aldeed: @jperl Good catch! Actually maybe it is best to have no maximum when going up? I'm thinking about a very complicated SVG inside a button, where the click might actually fire on some deeply nested SVG shape. We wouldn't want to stop until we come to the button, even if it's 100 elements up. The max depth is primarily useful when going down, where we could have lots of big branches and we are actually adding elements to the group.

aldeed: Changed it to check only on the way down in https://github.com/qawolf/qawolf/pull/832/commits/1414c64b0ed819de5a94b06f652e9816481421db. Let me know if you disagree.

aldeed:
@jperl Looks like the logic for clicking an `a` group within a `button` group fails on CI webkit only. I'll see if I can figure out why locally.

src/web/element.ts
jperl: I agree with your approach 👍

aldeed:
The test failure is because `computedStyle.cursor` is `auto` rather than `pointer` for `a > span` in webkit only. This is an example of where checking the cursor style isn't a great way to determine whether something is "clickable". I'll see if I can easily improve `isClickable`, or if not I'll just change the test HTML for now to avoid this webkit quirk.

aldeed:
@jperl Sandbox will need another release, but hopefully all tests should pass now. I realized that the `computedStyle.cursor === 'pointer'` check isn't really necessary anymore, so I deleted `isClickable` and I'm just using `isVisible` directly. Now a "click group" is defined solely by the fact that a click occurred in some descendant of a `button`, `a`, `label`, `role="button"`, or any button-type input.

jperl:
> @jperl Sandbox will need another release, but hopefully all tests should pass now. I realized that the `computedStyle.cursor === 'pointer'` check isn't really necessary anymore, so I deleted `isClickable` and I'm just using `isVisible` directly. Now a "click group" is defined solely by the fact that a click occurred in some descendant of a `button`, `a`, `label`, `role="button"`, or any button-type input.

I don't know how I feel about this. If you click a `div` won't it keep crawling to the `html` element even if it's parents aren't clickable. I think the `pointer` heuristic is imporant.

aldeed:
It does go all the way up to html for a div but since it never finds a click group top element, it treats it as not grouped. So there is a slight performance hit, but no difference in selector generation. But my gut says the isClickable style check will have numerous other problems as this is used more widely.

Adding likelyTopClickableElement check wouldn't help because the span would fail that too.

jperl:
> It does go all the way up to html for a div but since it never finds a click group top element, it treats it as not grouped. 

Then it will never include any descendants, but I guess that is ok.

jperl (approved):
❤️

PR #834, opened 2020-09-09
feat: ignore some dynamic-value attributes

### LINKED ISSUE #829: Ignore attributes with dynamic values
https://www.musement.com/us/ 

Screen Shot 2020-09-08 at 8 36 07 AM

`[data-test="autocomplete"] [name="3kty8fhtskl"]`

We should ignore any attributes with dynamic values. https://github.com/qawolf/qawolf/blob/d4e230f00c0e453ae6510ed8a6f93c18244b447f/src/web/cues.ts#L240

aldeed:
Resolves #829 

When building cues for an element, attributes that have a value that is either an empty string or appears to be dynamic (machine-generated) are now not used as cues. There are two exceptions: `href` and `src` which would frequently have dynamic values as part of the URL. Other attributes can be made exempt if you can think of reasons, or we could try to make `isDynamic` smarter.

Note that currently a half-dynamic value such as `input-d8sdgjk3` will be used. This is up for debate.

**Update**

Above is the original description, but here's how this actually works at the point of merging it.

- I left `isDynamic` pretty much the same
- But instead of running attribute values through `isDynamic`, I created a new function `getValueMatches` that attempts to find all the static portions of the value string that we can match on
- And then it will use the appropriate CSS selector operator based on what it finds (starts with, ends with, contains, or equals)
- And it still skips this for all `DYNAMIC_VALUE_OK` attributes, and now also for all preferred attributes.
- Also, rather than trying to specifically list HTML tags in the `words` list, I added the `html-tags` package and merged in the array provided by that package, which fixed some existing tests that were otherwise failing

So for example you now get selectors like this:

```js
await page.click('[name="nonDynamicInput"]'); // name="nonDynamicInput"
await page.click('[name^="input-"]'); // name="input-bu32879fDi"
await page.click('[name$="-check"]'); // name="bu32879fDi-check"
await page.click('[name*="-blue-"]'); // name="f89ndrn4-blue-5"
```

### DIFF
--- a/src/web/cues.ts
+++ b/src/web/cues.ts
@@ -1,8 +1,8 @@
 import { getAttribute } from './attribute';
 import { getElementText } from './selectorEngine';
-import { isDynamic } from './isDynamic';
+import { getValueMatchSelector, isDynamic } from './isDynamic';
 
-const DYNAMIC_VALUE_OK_ATTRIBUTES = ['href', 'src'];
+const DYNAMIC_VALUE_OK_ATTRIBUTES = ['href', 'src', 'value'];
 
 export type Cue = {
   level: number; // 0 is target, 1 is parent, etc.
@@ -19,6 +19,7 @@ export type BuildCues = {
 
 type CueTypeConfig = {
   elements: string[];
+  isPreferred?: boolean;
   penalty: number;
 };
 
@@ -129,6 +130,7 @@ export const getCueTypesConfig = (attributes: string[]): CueTypesConfig => {
   attributes.forEach((attribute) => {
     preferredAttributes[attribute] = {
       elements: ['*'],
+      isPreferred: true,
       penalty: 0,
     };
   }, {});
@@ -183,7 +185,7 @@ export const buildCuesForElement = ({
   }
 
   const cues: Cue[] = Object.keys(cueTypesConfig).reduce((list, cueType) => {
-    const { elements, penalty } = cueTypesConfig[cueType];
+    const { elements, isPreferred, penalty } = cueTypesConfig[cueType];
 
     // First find out whether this cue type is relevant for this element
     if (!elements.some((selector: string) => element.matches(selector))) {
@@ -244,13 +246,23 @@ export const buildCuesForElement = ({
         });
         if (attributeValuePair) {
           const { name, value } = attributeValuePair;
-          if (value.length && (DYNAMIC_VALUE_OK_ATTRIBUTES.includes(name) || !isDynamic(value))) {
+          if (value.length && (isPreferred || DYNAMIC_VALUE_OK_ATTRIBUTES.includes(name))) {
             list.push({
               level,
               penalty,
               type: 'attribute',
               value: `[${name}="${value}"]`,
             });
+          } else {
+            const { match, operator } = getValueMatchSelector(value) || {};
+            if (match) {
+              list.push({
+                level,
+                penalty,
+                type: 'attribute',
+                value: `[${name}${operator}"${match}"]`,
+              });
+            }
           }
         }
         break;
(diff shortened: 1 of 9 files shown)


src/web/cues.ts
```diff
@@ -2,6 +2,8 @@ import { getAttribute } from './attribute';
 import { getElementText } from './selectorEngine';
 import { isDynamic } from './isDynamic';
 
+const DYNAMIC_VALUE_OK_ATTRIBUTES = ['href', 'src'];
```
jperl, line 5: We should include "value" too.

jperl (approved):
Just need to fix the selector tests then 👍

jperl:
@aldeed Seems like you were able to "just" fix it 😜. Love your approach. Everything looks good, at a quick glance I find `getValueMatchSelector` hard to understand.

aldeed:
Yes, it's very hard to understand. I wrote it a simpler way first, but I think it would have been pretty inefficient. I'll add more comments and think about whether anything could be simplified.

aldeed:
@jperl I rewrote the new function to return multiple matches, which then become multiple attribute value cues. I realized this is better because then cue optimization logic can pick what's best. This had the added benefit of making the code easier to understand, but I also added a bunch of comments explaining the flow.

Sandbox needs to be published, but 🤞 this will be all good now.

jperl:
@aldeed The refactor & comments are great 💯 I was able to follow it. Publishing sandbox now.

jperl:
There is just one remaining issue, we run into selectors with [multiple split characters](https://github.com/qawolf/qawolf/blob/main/test/web/isDynamic.test.ts#L44). I updated [this test](https://github.com/qawolf/qawolf/pull/834/commits/f0e573749d37a81e12e71c5dbe3628cf569b23cf) and it shows there is a current bug with that.

I think the behavior we want is that if there are multiple, the `ValueMatch` will only use the one right before/after.

aldeed:
I fixed that issue, but I made it include all split characters rather than just the immediate ones, partially because that was easiest and partially because it seems to me that more characters would be a more likely match? I don't have strong feelings about that, so I can easily keep just the first before/after if you're sure that's best.

src/web/isDynamic.ts
```diff
@@ -182,12 +182,14 @@ export const getValueMatches = (
...
-      if (SPLIT_CHARACTERS.includes(lastCharOfPreviousBlock)) {
-        currentStaticBlock = lastCharOfPreviousBlock;
-      } else {
-        currentStaticBlock = '';
+      // block string, and potentially add all preceding split characters to it.
+      currentStaticBlock = '';
+      let backwardPosition = currentPosition - 1;
+      let previousCharacter = value[backwardPosition];
+      while (SPLIT_CHARACTERS.includes(previousCharacter)) {
+        if (tokenType === 'static') currentStaticBlock = previousCharacter + currentStaticBlock;
```
jperl, line 190: Is this if statement necessary? The tokenType should be static https://github.com/qawolf/qawolf/pull/834/commits/3c94c7f510957ca5d1856f7b59ee6f19230ffcc7#diff-457f25ee3f7d5025b30b8dfb3d70da85R183

jperl:
Awesome work. Merge it! 🚢

src/web/cues.ts
```diff
@@ -1,6 +1,8 @@
 import { getAttribute } from './attribute';
 import { getElementText } from './selectorEngine';
-import { isDynamic } from './isDynamic';
+import { getValueMatches, isDynamic } from './isDynamic';
+
+const DYNAMIC_VALUE_OK_ATTRIBUTES = ['href', 'src', 'value'];
```
jperl, line 5: let's add placeholder to this

jperl:
2 small comments. Then merge it 🚢 . Awesome work 🎉

src/web/isDynamic.ts
aldeed: Good catch. Fixed.

src/web/cues.ts
aldeed: Done

PR #836, opened 2020-09-11
feat: record and generate nav steps

### LINKED ISSUE #543: 📑Create code for enter url / open new tab

aldeed:
Resolves #543 and #438 

## Changes
- BREAKING `npx qawolf create`, without `url` argument, will now start a browser with a blank tab, which you can then load any address to start recording steps. Previously it would use `http://example.org` URL when you didn't specify one.
  - If you do `npx qawolf create` without `--name`, the test file name will be `qawolf.test.js/ts`
- The default test file template no longer includes the `page = await context.newPage();` line or the `page.goto` line for the initial URL. These are added by the existing page variable logic instead, as soon as there is a `page.goto` step to generate. If you start recording without specifying a URL, then this happens after you enter a valid URL in the address bar on the first blank tab.
  - Due to these changes, the `page` variable is no longer defined with file-level scope, but the `context` variable now is.
- The following navigation changes are captured and added to the generated code file:
  - Click Back
  - Click Reload
  - Click New Tab, and then enter a URL in that tab
- Clicking Forward is implemented but currently commented out due to one unfixable edge case bug. The bug:
  - When you click a link, then click back, then click the same link, we can't tell the difference between this and clicking Forward because the browser recognizes that you've already been to the link and simply moves forward in history entries, identical to clicking Forward. This results in both a `click` (for the link) and a `goForward` being generated in the test file, which in some cases would cause the test to fail when running it.

## Testing
1. Try `qawolf create` without URL. When it opens, go to any address. Stop recording and ensure that the generated test runs as expected.
2. Try `qawolf create` with a URL. Initial `page` lines should be added to your test. Stop recording and ensure that the generated test runs as expected.
3. When creating a test, open a new tab and type an address. Switch between tabs doing things on each. Verify that the test can be run and reproduces what you did.
4. When creating a test, click the Reload browser button. Verify that a `page.reload` line is added to the generated test for the correct page.
5. When creating a test, click the Back browser button. Verify that a `page.goBack` line is added to the generated test for the correct page.
6. Verify existing feature still works: `waitForPage` line is added whenever you do an action on a page that was opened by clicking a link that targeted a new tab.
7. Verify existing feature still works: `bringToFront` lines are added whenever you switch between doing actions on different pages (if `waitForPage` wasn't called)


aldeed:
@jperl I think this is ready. There is one test still failing for me locally in `create.e2e.test.ts`, but I'm not sure why. Maybe timing? We'll see if it fails on CI.

src/utils/context/waitForPage.ts
```diff
@@ -13,21 +12,9 @@ export const waitForPage = async (
   index: number,
   options: WaitForPageOptions = {},
 ): Promise<IndexedPage> => {
-  // index pages if they are not yet
-  await indexPages(context);
+  const page = await getPageAtIndex(context, index, { timeout: options.timeout });
```
jperl, line 15: Any reason to break this out?

jperl:
I noticed one bug. If you 

- `npx qawolf create`
- type google.com in the address bar
- click the back button
- type news.google.com in the address bar

```
const page = await context.newPage();
await page.goto("https://www.google.com/?gws_rd=ssl", { waitUntil: "domcontentloaded" });
await page.goBack();

// it does not record the second goto here
```

This only happens for the beginning route. Unless it is trivial to fix, I think it is OK to ignore this corner case.

jperl (approved):

aldeed:
Might be an easy fix. I'll check it out.

aldeed:
@jperl That bug should be fixed in https://github.com/qawolf/qawolf/pull/836/commits/bfb1cb947ee041e450d745d935207ac1f8e9c659. Added a test for it.

jperl:
@aldeed I see that, lgtm! 🚢

PR #840, opened 2020-09-18
Refactor buildSelector to be speedier

aldeed:
## Performance Changes
- Fixes an issue that prevented cached selectors from being used when the chosen selector is a part of a click group and not the original target.
    - On a click, this roughly cuts in half the amount of time it takes to add the click step because the `click` event can reuse the selector built for the nearly simultaneous `mousedown`
- Reduces the number of loops and consolidates some handling for `buildCues` and `optimizeCues`
    - One big help was: in `findBestCueGroup`, now uses `getElementMatchingSelectorParts` and `targetGroup.includes` rather than running `isMatch` multiple times.
    - On a click, this roughly reduces the amount of time it takes to add the click step by a factor equal to the number of elements in the click group. For example, a click group that has 4 elements and previously took 4 seconds might now take 1 second.

## Notes
Of the remaining time, most is spent in the `createTextSelector` function (getting element text). We will look at improving that in the future if necessary.

## Testing
Verify that clicks on click groups don't hang or hang for much less time, but that otherwise nothing about the generated code has changed.

### DIFF
--- a/src/web/element.ts
+++ b/src/web/element.ts
@@ -2,6 +2,16 @@ import { getXpath } from './serialize';
 
 type OnElementAddedToGroupFn = (element: HTMLElement, depth: number) => void;
 
+type TraverseClickableElementsInput = {
+  ancestorChain?: string[];
+  depth?: number;
+  direction?: 'up' | 'down';
+  element: HTMLElement;
+  group: HTMLElement[];
+  maxDepth?: number;
+  onElementAddedToGroup?: OnElementAddedToGroupFn;
+};
+
 const BUTTON_INPUT_TYPES = ['button', 'image', 'reset', 'submit'];
 const CLICK_GROUP_ELEMENTS = ['a', 'button', 'label'];
 const MAX_CLICKABLE_ELEMENT_TRAVERSE_DEPTH = 10;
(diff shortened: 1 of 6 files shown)


src/web/element.ts
```diff
@@ -42,6 +44,7 @@ export const isLikelyTopOfClickGroup = (element: HTMLElement): boolean => {
 const traverseClickableElements = (
```
jperl, line 44: This is a lot of positional arguments, perhaps an options object would be better.

jperl:
Seems like we still have the opportunity to short-circuit by ranking cue groups, and trying those in order of penalty. However this is fast enough for now w/ my testing so merge it 👍. At some point we will probably want to do a nice big refactor to clean up this logic.

src/web/optimizeCues.ts
```diff
@@ -228,36 +237,44 @@ export const findBestCueGroup = (
...
+
+  const cueSets = buildCueSets(cues);
+
+  for (let index = 0; index < cueSets.length; index++) {
+    // Only use the first 50 cue sets (there should never be this many, usually just ~2-3)
+    if (index > 49) break;
+
+    const cueSet = cueSets[index];
+
+    // Trim down the cue group to 10 if possible
```
jperl, line 261: Plz update this to 

```
// Trim down the cue group before exhaustively trying the combinations
````

Also I think this should be 8 instead of 5.

File: src/web/optimizeCues.ts [hunk 3]
```diff
@@ -228,36 +237,44 @@ export const findBestCueGroup = (
...
+    // 10 cues, samples of 5 is ~700 combinations which took ~20ms on my machine
+    const cueGroup = trimExcessCues(cueSet, target, 5);
+
+    // Skip if we cannot trim the group to <= 16 cues (this should rarely happen)
```
jperl, line 265: We should change it from 16 -> 10 and this should be.

```
// Skip if we cannot trim the group (this should rarely happen)
// Then exhaustively try all combinations of 3
```

PR #854, opened 2020-11-29
docs: update install.mdx

gomesalexandre:
- "the" CLI doesn't make sense, as we don't specify _which_ CLI is it. Is it the QA Wolf CLI? Your system's CLI e.g bash or zsh? 
- Using "a" CLI, it's clear what we mean: the QA Wolf CLI
- On line 29 however, it makes perfect sense as we specify _which._


jperl (approved):
Thanks!

PR #904, opened 2021-01-15
Support case sensitive URLs and URLs with colons

### LINKED ISSUE #900: Support case sensitive URLs
## Bug description

If you paste a case sensitive URL in the input, when the code is generated it is changed.

## Steps to reproduce

1. Paste a case sensitive URL into the create modal
2. Notice code has a different URL

### LINKED ISSUE #901: Support colon on URL
## Bug description

Screen Shot 2021-01-14 at 14 23 17

Colon `:` is a [safe character for URL,](https://stackoverflow.com/questions/2053132/is-a-colon-safe-for-friendly-url-use) but QA Wolf doesn't accept it.

## Steps to reproduce

1. Click "Create test"
2. On modal window, type `http://en.wikipedia.org/wiki/Template:Welcome`
3. Submit

flaurida:
* resolves #900 
* resolves #901

### DIFF
--- a/app/lib/helpers.ts
+++ b/app/lib/helpers.ts
@@ -100,14 +100,13 @@ export const isValidURL = (url: string): boolean => {
 export const parseUrl = (url: string): string => {
   if (!url) return DEFAULT_URL;
 
-  const parsed = url.toLowerCase();
-  if (parsed.includes("http://") || parsed.includes("https://")) return parsed;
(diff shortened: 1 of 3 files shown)


jperl:
This is that rare piece of front end code that should have unit tests

jperl (approved):

PR #929, opened 2021-01-17
Reusable helpers

### LINKED ISSUE #917: Reuse helpers
- [ ] implement
- [ ] document

We have heard multiple requests to reuse functions across tests. https://github.com/qawolf/qawolf/discussions/909

- to share login code
- to accept a cookie popup
- to close a modal

**Proposed Solution**

A helpers/functions/utils file shared across all tests, that is injected before the test code.

Screen Shot 2021-01-17 at 12 56 46 PM

You can use the helpers in the tests.

Screen Shot 2021-01-17 at 12 57 54 PM

➕ This matches the same developer workflow you would expect. Closing a modal makes more sense as helper method than a test.
➕ Easier to implement since we don't need to parse the linked tests.
➖ One helper file might get long. For most use cases it should be fine.

**Alternate Solution**

Ability to nest tests. IE call a test from another test.

flaurida:
* resolves #917


jperl:
Updating our tests to using helpers should suffice for testing them, but we will need another end-to-end test for updating the helpers.

jperl (approved):

Issue #912, opened 2021-01-15
Browser doesn't connect sometimes


flaurida:
## Bug description

Browser: Brave (ad blockers off, MacOS)

Infinite loading spinner. Apparently refresh does not work but a new tab did. Screenshots below

![image (5)](https://user-images.githubusercontent.com/6321017/104787474-db1a5880-574c-11eb-971a-03e83d18640a.png)
![image (4)](https://user-images.githubusercontent.com/6321017/104787485-dd7cb280-574c-11eb-99bb-eafeb8a9c7bd.png)

## Steps to reproduce

Run test, ignore it for some time, try to run again

jperl:
This has been drastically improved. Still on the lookout for client errors.

jperl:
This seems to be mostly fixed. Will reopen with new reports.

PR #931, opened 2021-01-17
Do not mark test as pending if it already assigned

jperl:
- merge findRunner into one query to reduce db queries
- do not mark runner_requested_at when there is a runner assigned (this can happen during a concurrent assignment)

Relates to #912

### DIFF
--- a/app/server/models/runner.ts
+++ b/app/server/models/runner.ts
@@ -234,10 +234,23 @@ export const findRunner = async (
   if (!include_deleted) filter.deleted_at = null;
 
   if (id) filter.id = id;
-  if (run_id !== undefined) filter.run_id = run_id;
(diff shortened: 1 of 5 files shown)


app/server/models/runner.ts
```diff
@@ -234,10 +234,23 @@ export const findRunner = async (
...
-  let query = (trx || db)("runners").select("*").from("runners").where(filter);
+  let query = (trx || db)("runners").select("*").from("runners");
+
+  if (run_id && test_id) {
+    // allow either
+    query = query
+      .where(function () {
+        this.where({ run_id }).orWhere({ test_id });
+      })
+      // prefer the run's runner
```
jperl, line 246: obviously this will change with the new ui

flaurida (approved):

PR #958, opened 2021-01-22
Sentry fixes

flaurida:
* addresses issue where Sentry user not initialized if already logged in
* fixes run id missing from logs

### DIFF
--- a/app/components/UserContext.tsx
+++ b/app/components/UserContext.tsx
@@ -1,4 +1,5 @@
 import { createContext, FC, useContext, useEffect } from "react";
+import { updateIntercomUser } from "../hooks/intercom";
 
 import { useIdentifyPostHog } from "../hooks/postHog";
 import { useCurrentUser } from "../hooks/queries";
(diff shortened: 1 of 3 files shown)


app/hooks/mutations.ts
```diff
@@ -298,8 +295,6 @@ const handleAuthenticatedUser = ({
 
   localStorage.setItem(JWT_KEY, access_token);
   updateCurrentUser(user);
-  updateIntercomUser(user.email);
```
flaurida, line 301: moved this to the current user hook so it will run any time user updated, not just on log in/sign up

app/components/UserContext.tsx
```diff
@@ -50,6 +52,13 @@ export const UserProvider: FC = ({ children }) => {
     state.setTeamId(fallbackTeamId);
   }, [teamId, user]);
 
+  useEffect(() => {
+    if (!user) return;
```
jperl, line 56: Would it make sense for this part to handle logout too?

jperl (approved):

PR #971, opened 2021-01-29
feat: (experiment) embed code line chapters in video files

aldeed:
Disclaimer: This may not work correctly yet (specifically, we need to see how well the chapters actually align with test steps across many varied test runs).

This updates the `VideoCapture` service and the `VideoArtifactsHook` to keep track of step start and end times during a run, and then save that information as chapter metadata in the video file that gets uploaded. When you download the video file and play it with any player that supports chapters, you should then be able to skip directly to any step of the test run using the chapter menu.


jperl:
will be replaced

PR #972, opened 2021-01-29
Fix scrolling of environment modals on small screen heights

flaurida:


flaurida:
doing a different approach

PR #984, opened 2021-02-02
only expire runs in progress

jperl:
- only run in eastus2 for now


app/test/server/models/runner.test.ts
```diff
@@ -343,6 +343,13 @@ describe("findPendingTestOrRunId", () => {
     expect(pending).toEqual(null);
   });
 
+  it("does not returns a run if the location is not eastus2", async () => {
```
flaurida, line 346: typo

flaurida (changes requested):

PR #1005, opened 2021-02-09
Remove alert settings from trigger

flaurida:
* merge after #1004


app/test/server/services/alert/send.test.ts
```diff
@@ -118,16 +118,6 @@ describe("sendAlert", () => {
 
     await db("teams").update({ alert_only_on_failure: false });
   });
-
-  it("does not send alerts per team settings", async () => {
```
flaurida, line 122: no longer allowed to have no alerts

jperl:
Resolves #1007

PR #1018, opened 2021-02-11
Update triggers UI

### LINKED ISSUE #936: Triggers v2
**Goal**
Rather than adding tests to one or more groups to determine when the test should run and with which environment variables, assign triggers—when + environment—to tests as tags.

This is a change in the conceptual model of triggering tests to run, not necessarily new functionality. This enhancement should encourage the action of triggering new tests to run, and reduce the UI clutter/confusion of seeing the same test in multiple lists. Users will now be able to group their tests for organization and discoverability (i.e. by feature/functionality)

**Details**
- Triggers are tags that can be added to tests, instead of groups that tests are added to
  - Groups now become a mechanism to organize tests for better discoverability
- Provide a dedicated view to manage triggers
- Make it easy to do a few default triggers—"Hourly", "Daily"—and allow users to edit them and/or create new triggers
- When creating or editing a test, allow users to add triggers to the test
- When adding a trigger, allow users to create new triggers
- Triggers include a name, when (time-based or on deployment), and an environment

flaurida:
* resolves #936


jperl:
This PR is

jperl (approved):

PR #1023, opened 2021-02-12
Collect test metadata

jperl:
replaces #971


runner/src/environment/Run.ts
```diff
@@ -83,6 +91,8 @@ export class Run extends EventEmitter {
     this._progress.completed_at = new Date().toISOString();
     this._emitProgress();
 
+    this.off("runprogress", handleHookProgress);
```
jperl, line 94: off is necessary since run can be called multiple times

PR #1039, opened 2021-02-16
Include suite environment when editing test

### LINKED ISSUE #1040: Rerun test against proper deployment url
**Problem**

When my deployment test fails, I want to re-run it against the same deployment url.

flaurida:
Resolves #1040


jperl (approved):
Update our e2e test then 👍

PR #1042, opened 2021-02-16
Test receiving emails

flaurida:


app/server/api/sendgrid.ts
```diff
@@ -0,0 +1,71 @@
...
+  res: NextApiResponse,
+  { db, logger }: ModelOptions
+): Promise<void> => {
+  const log = logger.prefix("handleSendGridRequest");
+
+  try {
+    verifyRequest(req.url, logger);
+    const { from, html, subject, text, to } = await buildEmailFields(req);
+
+    await db.transaction(async (trx) => {
```
jperl, line 56: I don't think this transaction is necessary

jperl (approved):

PR #1046, opened 2021-02-16
Increase runner resolution

### LINKED ISSUE #1045: Larger devices are cut off

jperl:
Resolves #1045


jperl:
This solution does not work well for all device types. We should dynamically resize the browser insttead

PR #1050, opened 2021-02-18
Use email sent date to mark when email created

flaurida:


app/server/api/sendgrid.ts
```diff
@@ -29,6 +30,12 @@ export const buildEmailFields = async (
   });
 };
 
+export const buildSendDate = (headers: string): string => {
+  const timestamp = headers.split("Date:")[1]?.split("\n")[0];
+
+  return new Date(timestamp).toISOString();
```
jperl, line 36: check it is valid otherwise fallback to now

jperl (approved):

PR #1061, opened 2021-02-22
Minor docs tweaks

jperl:


app/components/Docs/Layout/Section/SectionLinks.tsx
```diff
@@ -41,7 +41,6 @@ const StyledSectionLinks = styled(SectionLinks)`
   // each link is 40 px tall including margin
   height: ${({ docs }) => `${docs.length * 40}px`};
```
flaurida, line 42: this can probably go away entirely if we aren't animating

PR #1064, opened 2021-02-22
Retry runs with ERR_CONNECTION_REFUSED error

### LINKED ISSUE #1062: Retry ERR_CONNECTION_REFUSED errors
Every ~1/5000 runs, the Azure container might have intermittent connectivity issues (bad rack, co-tenants, etc).

Even though it's rare, it is annoying to find a run failed a navigation if it has nothing to do with your site.

Store and retry ERR_CONNECTION_REFUSED once. We can expand this list over time.

jperl:
Resolves #1062

### DIFF
--- a/app/server/resolvers/runner.ts
+++ b/app/server/resolvers/runner.ts
@@ -168,8 +168,8 @@ export const updateRunnerResolver = async (
 
   let runId = runner.run_id;
 
-  // assign the runner if there is a pending test or run
-  if (!runner.run_id && !runner.test_id) {
+  // assign the runner if it's session has not been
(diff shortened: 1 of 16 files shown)


app/server/resolvers/runner.ts
```diff
@@ -168,8 +168,8 @@ export const updateRunnerResolver = async (
 
   let runId = runner.run_id;
 
-  // assign the runner if there is a pending test or run
-  if (!runner.run_id && !runner.test_id) {
+  // assign the runner if it's session has not been
```
flaurida, line 171: its

flaurida (approved):

PR #1073, opened 2021-02-23
pool: consider incomplete runs

### LINKED ISSUE #1071: Runner pool is too conservative
We are not considering assigned but incomplete runs.

jperl:
Resolves #1071 

- Include in progress tests and runs in pool calculations
- Fix a potential race condition where an assigned runners could be deleted during rebalancing
- Retry in progress runs for deleted unhealthy runners once
- Alert us to unhealthy runners
- Add db constraint to ensure tests are not pending after deleted
- Consolidated reset runner logic (delete, expire, restart)

### DIFF
--- a/app/server/models/test.ts
+++ b/app/server/models/test.ts
@@ -272,7 +272,7 @@ export const deleteTests = async (
 
   const tests = await db.transaction(async (trx) => {
     const tests = await trx.select("*").from("tests").whereIn("id", ids);
-    const updates = { deleted_at: minutesFromNow() };
+    const updates = { deleted_at: minutesFromNow(), runner_requested_at: null };
     await trx("tests").update(updates).whereIn("id", ids);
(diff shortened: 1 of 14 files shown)


app/server/models/test.ts
```diff
@@ -61,18 +61,20 @@ export const buildTestName = async (
...
   { db, logger }: ModelOptions
 ): Promise<LocationCount[]> => {
-  const log = logger.prefix("countPendingTests");
+  const log = logger.prefix("countIncompleteTests");
 
   const countQuery = `SELECT COUNT(s.*), s.location
   FROM (
     SELECT coalesce(runner_locations::json->> 0, ?) as location
     FROM tests
-    WHERE runner_requested_at IS NOT NULL
```
jperl, line 74: we don't need the deleted_at clause anymore since we are constraining deleted runs to not have runner_requested_at

flaurida (approved):

PR #1093, opened 2021-02-26
Basic logs about email api key

flaurida:


jperl:
Will replace properly

PR #1099, opened 2021-02-27
Do not throw error if environment deleted in suite resolver

flaurida:


app/test/server/resolvers/suite.test.ts
```diff
@@ -221,18 +221,20 @@ describe("suiteResolver", () => {
...
+      ...suites[0],
+      environment_id: "deletedId",
+      trigger_id: null,
+    });
 
     const suite = await suiteResolver({}, { id: "suiteId" }, context);
 
     expect(suite).toEqual({
       ...suites[0],
-      environment_id: null,
```
jperl, line 235: shouldn't this be null for a deleted test

app/components/Modals/Environments/Environments/ConfirmDelete.tsx
```diff
@@ -20,10 +21,14 @@ export default function ConfirmDeleteEnvironment({
...
   onClose,
 }: Props): JSX.Element {
+  const { environmentId } = useContext(StateContext);
+
   const [error, setError] = useState("");
   const [name, setName] = useState("");
 
-  const [deleteEnvironment, { loading }] = useDeleteEnvironment();
+  const [deleteEnvironment, { loading }] = useDeleteEnvironment({
+    environmentId,
```
jperl, line 30: maybe this should be `currentEnvironmentId` or `stateEnvironmentId`. Otherwise I would think this is the environmentId used for deletion.

jperl (approved):

PR #1110, opened 2021-03-03
Filter trigger counts if group selected

flaurida:


app/components/Dashboard/Tests/Header/SelectTrigger.tsx
```diff
@@ -16,19 +16,26 @@ type Props = {
 const dividerProps = { margin: { vertical: "xxxsmall" } };
 const width = "320px";
 
+const buildBasePath = (groupId?: string): string => {
```
jperl, line 19: buildTestsPath

app/server/models/test_trigger.ts
```diff
@@ -105,14 +105,16 @@ export const findTestTriggersForTests = async (
   const log = logger.prefix("findTestTriggersForTests");
   log.debug(testIds);
 
+  const tests = await db("tests").whereIn("id", testIds);
```
jperl, line 108: I would only select the things you need (id, group_id) to make this faster.

PR #1128, opened 2021-03-09
Add mailing list option to sign up

flaurida:


app/components/Auth/EmailForm.tsx
```diff
@@ -36,14 +36,6 @@ export default function EmailForm({
   return (
     <Keyboard onEnter={handleClick}>
       <Box flex={false}>
-        <Text
```
flaurida, line 39: removed this to conserve real estate

PR #1174, opened 2021-03-28
Update API for Netlify plugin

flaurida:


app/graphql/mutations.ts
```diff
@@ -150,10 +150,12 @@ export const createTestMutation = gql`
 export const createTriggerMutation = gql`
   mutation createTrigger(
     $deployment_branches: String
-    $deployment_environment: DeploymentEnvironment
```
flaurida, line 153: graphql does not allow enum names with dashes, so `deploy-preview` a no go

app/components/Modals/Triggers/CreateOrEditTrigger/DeployFields.tsx
```diff
@@ -0,0 +1,87 @@
...
+  netlifyEvent,
+  setDeployBranches,
+  setDeployEnv,
+  setDeployIntegrationId,
+  setDeployProvider,
+  setNetlifyEvent,
+}: Props): JSX.Element {
+  // clear settings if switching providers
+  useEffect(() => {
+    if (deployProvider === "netlify") {
```
jperl, line 43: lets do just preview and production

app/components/Modals/Triggers/helpers.ts
```diff
@@ -71,12 +79,20 @@ export const buildTriggerFields = ({
   if (mode === "deployment") {
     return {
       ...constantFields,
-      deployment_branches: deployBranches || null,
-      deployment_environment: ["preview", "production"].includes(deployEnv)
+      deployment_branches:
+        deployBranches && deployProvider === "vercel" ? deployBranches : null,
+      deployment_environment: [
+        "deploy-preview",
```
jperl, line 85: this is removed?

app/components/shared/TriggerIcon.tsx
```diff
@@ -10,7 +10,7 @@ import Rocket from "./icons/Rocket";
 type Props = { trigger: ShortTrigger | null };
 
 const getTriggerIconComponent = (trigger: ShortTrigger): Icon => {
-  if (trigger.deployment_integration_id) return Rocket;
+  if (trigger.deployment_provider) return Rocket;
```
jperl, line 13: Should this be deployment_integration_id || deployment_provider

app/graphql/mutations.ts
jperl: That is fine this can support custom contexts too in the future

app/server/api/netlify/suites.ts
```diff
@@ -0,0 +1,204 @@
...
+  trigger: Trigger;
+};
+
+type CreateSuite = {
+  integration_id: string;
+  team_id: string;
+  trigger: Trigger;
+};
+
+class AuthenticationError extends Error {
```
jperl, line 23: This was moved to a shared file yes?

app/server/api/netlify/suites.ts
```diff
@@ -0,0 +1,204 @@
...
+    log.debug("skip: deployment_environment", deployment_environment);
+    return [];
+  }
+
+  const triggers = await findTriggersForNetlifyIntegration(
+    { deployment_environment, netlify_event, team_id: team.id },
+    options
+  );
+
+  const suites = await Promise.all(
```
jperl, line 124: should this be in a transaction?

app/server/api/netlify/suites.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  return suite_ids;
+};
+
+// errors example: https://stripe.com/docs/api/errors
+const findTeamForRequest = async (
+  req: NextApiRequest,
+  options: ModelOptions
+): Promise<Team> => {
+  const log = options.logger.prefix("findTeamForRequest");
```
jperl, line 149: Going forward we should re-use this for all apis that use an api key. Eventually (not now) we should refactor existing ones.

app/components/shared/TriggerIcon.tsx
flaurida: We should run a query to update the deployment triggers to have a `deployment_provider`

app/server/api/netlify/suites.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export const shouldCreateSuites = (
+  req: NextApiRequest,
+  { logger }: ModelOptions
+): boolean => {
+  const log = logger.prefix("shouldCreateSuites");
+
+  const { deployment_environment, is_pull_request, skip } = req.body;
+
+  if (skip && ["true", "t"].includes(skip.toLowerCase())) {
```
jperl, line 191: I would add 1 to this

jperl (approved):

app/server/api/netlify/suites.ts
flaurida: i am moving it to plugin, see https://github.com/qawolf/netlify-plugin-qawolf/pull/4

PR #1181, opened 2021-03-31
Only show complete label if complete in tutorial

flaurida:


app/components/Dashboard/GetStarted/Section/index.tsx
```diff
@@ -33,7 +33,9 @@ export default function Section({
   onToggleOpen,
   section,
 }: Props): JSX.Element {
-  const buttonLabel = completeButtonLabels[section] || buttonLabels[section];
+  const buttonLabel = isComplete
```
jperl, line 36: const buttonLabel = (isComplete && completeButtonLabels[section]) || buttonLabels[section];

should be the same

PR #1185, opened 2021-04-01
add user.intercom_hash

jperl:


app/graphql/queries.ts
```diff
@@ -21,6 +21,7 @@ import {
 export const currentUserQuery = gql`
   query currentUser {
     currentUser {
+      intercom_hash
```
flaurida, line 24: why not add to `UserFragment`?

jperl: It's not actually included everywhere only in CurrentUser query

flaurida (approved):
1 minor comment then lgtm

app/graphql/queries.ts
flaurida: ok

PR #1190, opened 2021-04-02
Allow subscribing to plan

flaurida:


app/components/Dashboard/Settings/Plan.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const {
+    isLoading: isCheckoutLoading,
+    onClick: onCheckoutClick,
+  } = useOnStripeCheckout();
+  const {
+    isLoading: isPortalLoading,
+    onClick: onPortalClick,
+  } = useOnStripePortal();
+
+  const isSubscribed = team.plan === "business";
```
jperl, line 27: should this be if team.plan !== "free"?

app/lib/stripe.ts
```diff
@@ -0,0 +1,23 @@
...
+
+class Stripe {
+  private stripe: StripeType | null = null;
+
+  constructor() {
+    this.load();
+  }
+
+  async redirectToCheckout(sessionId: string): Promise<void> {
+    if (!this.stripe) await this.load();
```
jperl, line 11: I would put `if (this.stripe) return` inside load instead

app/lib/stripe.ts
```diff
@@ -0,0 +1,23 @@
+import { loadStripe, Stripe as StripeType } from "@stripe/stripe-js";
+
+class Stripe {
+  private stripe: StripeType | null = null;
+
+  constructor() {
+    this.load();
```
jperl, line 7: I would remove this

app/lib/stripe.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  async redirectToCheckout(sessionId: string): Promise<void> {
+    if (!this.stripe) await this.load();
+
+    this.stripe.redirectToCheckout({ sessionId });
+  }
+
+  private async load(): Promise<void> {
+    this.stripe = await loadStripe(
```
jperl, line 17: You should store the promise and either return that or create a new one.

app/lib/stripe.ts
jperl, line 11: I would move this check to load() and short circuit if loaded

jperl (approved):

PR #1214, opened 2021-04-11
Track run usage

flaurida:


app/server/jobs/updateSegmentTeams.ts
```diff
@@ -1,24 +1,83 @@
...
+
+export const syncTeam = async (
+  team: Team,
+  options: ModelOptions
+): Promise<Team> => {
+  const log = options.logger.prefix("syncTeam");
+  const timestamp = minutesFromNow();
+
+  const shouldRenew =
+    team.plan === "free" && team.renewed_at < daysFromNow(-30);
```
jperl, line 20: Is this safe to do as a string comparison?

flaurida: what do you mean safe?

flaurida: oh for the date? good call let me write a test

app/graphql/queries.ts
```diff
@@ -78,6 +78,12 @@ export const onboardingQuery = gql`
   }
 `;
 
+export const runCountQuery = gql`
+  query runCount($team_id: ID!) {
+    runCount(team_id: $team_id)
```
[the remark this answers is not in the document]
flaurida: no since it just returns an int you don't need to query subfields

app/server/jobs/updateSegmentTeams.ts
```diff
@@ -1,24 +1,94 @@
...
+    ? { limit_reached_at: null, renewed_at: timestamp }
+    : {};
+
+  return updateTeam(
+    { id: team.id, ...renewFields, last_synced_at: timestamp },
+    options
+  );
+};
+
+export const updateSegmentTeam = async (
```
jperl, line 44: I think this job/file should be renamed to sync team

app/server/jobs/updateSegmentTeams.ts
```diff
@@ -1,24 +1,94 @@
...
+
+  const isExpired = new Date(team.renewed_at) < new Date(daysFromNow(-30));
+  const renew = team.plan === "free" && isExpired;
+
+  log.debug("should renew?", renew);
+
+  return renew;
+};
+
+export const syncTeam = async (
```
jperl, line 28: updateTeamSyncedAt?

app/server/jobs/updateSegmentTeams.ts
```diff
@@ -1,24 +1,94 @@
...
+  );
+};
+
+export const updateSegmentTeam = async (
+  originalTeam: Team,
+  options: ModelOptions
+): Promise<void> => {
+  const log = options.logger.prefix("updateSegmentTeam");
+
+  const team = await syncTeam(originalTeam, options);
```
jperl, line 50: This should be called after everything else? Otherwise the rest could fail but the team will show synced_at is still updated

app/server/jobs/updateSegmentTeams.ts
```diff
@@ -1,24 +1,94 @@
...
+
+  const runCount = await countRunsForTeam(team, options);
+  const testCounts = await countTestsForTeam(team.id, options);
+
+  const free_limit_reached = team.plan === "free" && runCount >= freePlanLimit;
+
+  if (free_limit_reached) {
+    await updateTeam(
```
jperl, line 58: Can this be combined into one `updateSyncTeam` function

app/server/jobs/updateSegmentTeams.ts
jperl: I'd leave a comment to why this happens before counting runs

jperl (approved):

PR #1221, opened 2021-04-14
Persist position in helpers and code tabs

### LINKED ISSUE #1220: Persist cursor position when switching between tabs in editor

flaurida:
* resolves #1220

### DIFF
--- a/app/components/Editor/Sidebar/CodeEditor/index.tsx
+++ b/app/components/Editor/Sidebar/CodeEditor/index.tsx
@@ -16,7 +16,13 @@ export default function CodeEditor({ onKeyDown }: Props): JSX.Element {
   const [editor, setEditor] = useState<Editor | null>(null);
   const [monaco, setMonaco] = useState<typeof monacoEditor | null>(null);
 
-  const { env, progress, onSelectionChange } = useContext(RunnerContext);
+  const {
+    codeLine,
+    env,
+    progress,
+    onCodeLineChange,
+    onSelectionChange,
+  } = useContext(RunnerContext);
   const { code, controller, hasWriteAccess, team } = useContext(TestContext);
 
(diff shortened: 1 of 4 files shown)


app/components/Editor/Sidebar/HelpersEditor.tsx
```diff
@@ -45,13 +45,18 @@ export default function HelpersEditor({ onKeyDown }: Props): JSX.Element {
...
+        editor.setPosition({ column: 1, lineNumber: helpersLine });
+      }
     }
 
     // update current team helpers version so it works in callback
     if (team.helpers_version > -1) {
       helpersVersionRef.current = team.helpers_version;
     }
-  }, [editor, team]);
+  }, [editor, helpersLine, team]);
```
flaurida, line 59: I don't want to re-run the hook every time this changes, safe to ignore `helpersLine` as a dependency?

app/components/Editor/Sidebar/CodeEditor/index.tsx
```diff
@@ -28,14 +34,24 @@ export default function CodeEditor({ onKeyDown }: Props): JSX.Element {
...
+    );
   };
 
   useEffect(() => {
     if (!controller || !editor) return;
 
     // set the initial editor code
     editor.setValue(controller.code);
+    // restore position
+    if (codeLine) {
```
flaurida, line 51: likewise not sure `codeLine` should be a dependency of `useEffect`

flaurida: (if it is it keeps setting the editor value)

jperl (changes requested):
Going to submit a different PR with a simpler approach, just don't remove the elements from the dom when switching tabs

PR #1224, opened 2021-04-14
Analytics

jperl:


app/server/resolvers/user.ts
```diff
@@ -103,11 +104,10 @@ const createUserWithTeam = async (
           { db: trx, logger }
         );
 
+    trackSegmentEvent({ event: "Signed Up", user });
```
flaurida, line 107: I think the Segment nomenclature would be `User signed up`

flaurida: make sure capitalization consistent with other events

flaurida (approved):
1 comment then lgtm

PR #1241, opened 2021-04-20
use chrome channel for chrome

jperl:
This feature from playwright 1.10 auto-handles the executablePath


flaurida (approved):
fix tests first then lgtm

PR #1257, opened 2021-04-27
add runner branch

jperl:


app/server/migrations/20210427114638_update_runners_tests_test_branch.ts
```diff
@@ -0,0 +1,41 @@
+import * as Knex from "knex";
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable("runners", (table) => {
+    table.string("test_branch");
```
flaurida, line 5: why not just `branch`

flaurida (approved):

app/server/migrations/20210427114638_update_runners_tests_test_branch.ts
jperl: Because its only the test branch, never the run branch

PR #1273, opened 2021-04-30
chore: proposed new develop scripts

aldeed:
This is a proposal for an easier way to run the projects in whatever configuration is necessary for local development. Marking as draft since I haven't thoroughly tested on various OS yet.

The commands are documented in CONTRIBUTING.md:
- is this where we want them? or in readme? docs? all?
- potentially also add database to docker-compose file or better document that setup


aldeed:
This was revised and merged in https://github.com/qawolf/qawolf/pull/1286. Closing

PR #1274, opened 2021-04-30
Sync helpers, code, name, path, and saved* versions

jperl:
Doing this with a shared naively versioned map built on top of socketio. Will definitely improve this next week.


app/components/Editor/Sidebar/CodeEditor/index.tsx
```diff
@@ -23,44 +23,22 @@ export default function CodeEditor({
   const [monaco, setMonaco] = useState<typeof monacoEditor | null>(null);
 
   const { env, progress, onSelectionChange } = useContext(RunnerContext);
-  const { code, controller, hasWriteAccess, team } = useContext(TestContext);
+  const { controller, code, hasWriteAccess, team } = useContext(TestContext);
```
flaurida, line 26: abc

flaurida (approved):

PR #1276, opened 2021-05-03
create bash CLI script for most common dev env tasks

jshimko:
The updates/changes...

- Dockerize the NextJS app
- set a dedicated location for `WORKDIR` in the runner build so that files aren't all mixed in with root OS dirs in the container
- updated entrypoint script for the runner container
- created a bash CLI script for most common dev env tasks

## How-to

Since the script help message is fairly self explanatory, here's the summary...

```
./wolf help


Usage:

 ./wolf [OPTION]

Options:

  init      Install deps in all projects, build the runner container, and run app migrations
  build     Build all containers (or specify a container to build with an additional argument)
  dev       Start postgres and runner containers, then start app in dev mode
  start     Run production builds in docker containers and tail logs app/runner logs
  stop      Stop all running containers
  clean     Clean up all containers, volumes, and node_modules (full reset)
```

## To Test

```sh
git clone  -b jer-dev-scripts --single-branch git@github.com:qawolf/qawolf.git dev-script-test
cd dev-script-test

# install all deps, create a .env, start postgres, and run migrations
./wolf init

# start postgres and runner, start app in dev mode
./wolf dev

# start postgres, runner, and app in containers in production mode
./wolf start

# stop all running containers
./wolf stop

# stop all running containers, delete docker volumes, delete all node_modules
# repo should be back to the same state as a fresh checkout (except your .env stays intact)
./wolf clean
```

Also note that `./wolf init` is pretty handy for any time you do a `git pull`.  It's a nice quick way to make sure all of your dependencies are up to date and the local runner image is a fresh build.

Let me know what you think!  Any feedback welcome.

### DIFF
--- a/wolf
+++ b/wolf
@@ -1,0 +1,149 @@
+#!/bin/bash
+
+set -e
+
+
+showHelp() {
+  printf "\nUsage: \n\n ./wolf [OPTION] \n"
+  printf "\nOptions:\n
+  init      Install deps in all projects, build the runner container, and run app migrations
+  build     Build all containers (or specify a container to build with an additional argument)
+  dev       Start postgres and runner containers, then start app in dev mode
+  start     Run production builds in docker containers and tail logs app/runner logs
+  stop      Stop all running containers
+  clean     Clean up all containers, volumes, and node_modules (full reset)
+  "
+}
+
+# if no arg was provided, show help
+if [ -z "$1" ] || [ "$1" == "help" ]; then
+  showHelp
+  exit 1
+fi
+
+
+# confirm docker 
+if ! command -v docker &> /dev/null; then
+  printf "\nDocker must be installed!\n"
+  printf "More info: https://docs.docker.com/engine/install/\n\n"
+  exit 1
+fi
+
+# confirm node/npm
+if ! command -v npm &> /dev/null; then
+  printf "\nNode and npm must be installed!\n"
+  printf "More info: https://nodejs.org/en/download/\n\n"
+  exit 1
+fi
+
+# ensure a .env file exists for the app and docker-compose file
+ensureDotenv() {
+  if [ ! -f ./app/.env ]; then
+    printf "\nCreating .env file...\n\n"
+    cp ./app/.env.example ./app/.env
+  fi
+}
+
+
+# install deps in all projects, run migrations, and build the runner container
+if [[ $1 == "init" ]]; then
+  ensureDotenv
+
+  # postgres
+  printf "\nStarting postgres...\n\n"
+  docker-compose up -d postgres
+
+  # app
+  printf "\nInstalling node_modules for app...\n\n"
+  cd app
+  npm i
+  
+  # qawolf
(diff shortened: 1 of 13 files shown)


aldeed (commented):
@jshimko This looks like a lot of good changes. I didn't try it out yet, but in reviewing the code, I actually don't see much overlap between this and my #1273 . We will probably want to merge both? Mine primarily deals with getting "hot reload" to work for both runner and recorder projects, which I don't see in yours. That was the main purpose of develop.ts.

So I'm thinking:
1. Merge this after I test it
2. I'll make some adjustments to my PR with your changes in mind.
3. Probably merge that, too

In the end I think we'd probably have something like multiple variations of the `wolf dev` command, which would allow you to run any combinations of the three apps in "dev mode", not just "app".

aldeed:
@jshimko One more thought... if you're touching the runner Dockerfile, maybe this is a good time to split the initial apt installation into a separate file / base image? The fact that an initial / fresh build takes 15+ minutes to do that package installation stuff seems completely unnecessary when we can just throw an image on DockerHub or GitHub and start from that.

jshimko:
> @jshimko This looks like a lot of good changes. I didn't try it out yet, but in reviewing the code, I actually don't see much overlap between this and my #1273 . We will probably want to merge both? Mine primarily deals with getting "hot reload" to work for both runner and recorder projects, which I don't see in yours. That was the main purpose of develop.ts.

This sounds good to me.  Yeah, I felt like they are different enough scripts that we could probably merge the ideas together somehow.  While bash was much simpler/faster to throw together, there are definitely some advantages to having Node (or Deno) API's available.  Particularly with that file watching stuff.  Trying to do that with bash is probably more trouble than it's worth considering file system stuff like that is a fairly solved problem in JS/Node.  Anyway yeah, let's keep both for now and figure out a way to merge all of the features into a single solution.

> if you're touching the runner Dockerfile, maybe this is a good time to split the initial apt installation into a separate file / base image?

Yeah, I had considered that.  Was trying to limit how much I changed in one PR, but now that you mention it, that'd be a fairly simple solution with no breaking changes to worry about.  I'll just create a base Dockerfile and get it pushed up to Docker Hub so we can just pull those layers down instead of building them every time.  I'll knock that out before we merge here.

wolf
```diff
@@ -72,6 +74,7 @@ if [[ $1 == "init" ]]; then
   printf "\nBuilding recorder script for runner container...\n\n"
   npm run build:recorder-script
   printf "\nBuilding runner image...\n\n"
+  docker build -f Dockerfile.base -t qawolf/runner-base:$DOCKER_BASE_VERSION .
```
aldeed, line 77: @jshimko Couldn't we let this be a separate command that also pushes to a remote repo, and then this command will auto-pull the base image rather than build it locally for each dev? In my experience downloading the image is like 10 times faster than building it, and the base image will rarely change.

jshimko: @aldeed Fair point. Updated it to do a pull on the `init` command.

PR #1281, opened 2021-05-05
add branch to api / cli

jperl:


qawolf/src/cli/testCommand.ts
```diff
@@ -7,8 +7,9 @@ export const buildTestCommand = (): program.Command => {
     .requiredOption("-t, --trigger <id>", "id of the trigger to run")
     .option("-e, --env <env>", "environment variables to pass to your tests")
     .option("-n, --no-wait", "do not wait for the tests finish running")
-    .action(async ({ env, trigger: triggerId, wait }) => {
-      await runTests({ env, triggerId, wait });
+    .option("-b, --branch", "the branch of tests to run")
```
flaurida, line 10: abc

jperl: They are ordered by likelihood of use

PR #1282, opened 2021-05-05
Migrate test groups to tags

### LINKED ISSUE #1260: 🏷️ Change groups to tags and improve UX for filtering tests
- Allow adding multiple tags to tests. People are using api triggers as a tag to workaround this.
- Allow filtering by a set of tags, "All" or "Any". Then triggering those manually.
- Replace API trigger by allowing tag names to be passed to the API. For the migration, we can migrate api triggers to tags. Then consider trigger_id to be the same as tag_id for backwards compatibility.

flaurida:
* resolves #1260


qawolf/src/cli/testCommand.ts
```diff
@@ -3,14 +3,26 @@ import runTests from "./runTests";
...
+      "comma separated list of tag names (example: Account,Checkout)"
+    )
     .option("-e, --env <env>", "environment variables to pass to your tests")
-    .option("-n, --no-wait", "do not wait for the tests finish running")
-    .option("-b, --branch <branch>", "the branch of tests to run")
-    .action(async ({ branch, env, trigger: triggerId, wait }) => {
-      await runTests({ branch, env, triggerId, wait });
-    });
+    .option(
+      "-n, --env-name <envName>",
```
jperl, line 15: I think we should just bump the major version for the CLI and change this to be `--environment <environmentName>` and `--variables <environmentVariables>`

qawolf/src/cli/testCommand.ts
```diff
@@ -3,14 +3,26 @@ import runTests from "./runTests";
...
+      "environment name to use when running your tests (example: Staging)"
+    )
+    .option("-s, --no-wait", "do not wait for the tests finish running")
+    .option("-b, --branch <branch>", "git branch of tests to run")
+    .option("-r, --trigger <id>", "deprecated: id of the trigger to run")
```
jperl, line 20: `-r` does not make sense to me. I would remove that short hand.

qawolf/src/cli/runTests.ts
```diff
@@ -26,7 +32,13 @@ const createSuite = async ({
       data: { id, url },
     } = await axios.post(
       suitesUrl,
-      { branch: branch || null, env, trigger_id: triggerId },
+      {
+        branch: branch || null,
+        env,
+        env_name: envName,
```
jperl, line 38: I think we should stay consistent and do `environment_name`, I also think we should add `variables` to the API and use that (but support env for backwards compat).

flaurida: if we're calling it `environment` should it just be `environment` without the name?

jperl: What if we allow `environment_id` in the future. Will that be confusing? I can see either way.

jperl: Yes I like that bette 👍

flaurida: but then for the api it is `tags` not `tag_names` 🤔 I prefer `environment` and also consistency between API and CLI since you can use either

app/components/Modals/CreateSuite.tsx
```diff
@@ -0,0 +1,92 @@
...
+    environmentId
+  );
+
+  const { data } = useEnvironments({ team_id: teamId }, { environmentId });
+  const environments = data?.environments || null;
+
+  const [createSuite, { loading }] = useCreateSuite();
+
+  // do not choose environment if no options
+  useEffect(() => {
```
jperl, line 36: when does this happen? I thought this modal only opens if there is > 1 environment?

flaurida: it does but this is just in case. for example if someone deleted the extra environment while modal was opening

qawolf/src/cli/runTests.ts
jperl: Ok I agree with you

app/components/Docs/ApiExample.tsx
```diff
@@ -0,0 +1,35 @@
...
+        style={{ whiteSpace: "break-spaces" }}
+      >{`curl -H "Authorization: ${apiKey}" -H "Content-Type: application/json" ${apiUrl}`}</CodeBlock>
+      <p>
+        You can also optionally specify environment variables, an environment
+        name, and test tags:
+      </p>
+      <CodeBlock
+        className="language-bash"
+        style={{ whiteSpace: "break-spaces" }}
+      >{`curl -H "Authorization: ${apiKey}" -H "Content-Type: application/json" ${apiUrl} -d '{"env": { "MY_VARIABLE": "secret" }, "env_name": "Staging", "tags": "Checkout,Sign up"}'`}</CodeBlock>
```
jperl, line 32: environment, tags, variables

app/components/Docs/CliExample.tsx
```diff
@@ -0,0 +1,32 @@
...
+        style={{ whiteSpace: "break-spaces" }}
+      >{`QAWOLF_API_KEY=${apiKey} npx qawolf test`}</CodeBlock>
+      <p>
+        You can also optionally specify environment variables, an environment
+        name, and test tags:
+      </p>
+      <CodeBlock
+        className="language-bash"
+        style={{ whiteSpace: "break-spaces" }}
+      >{`QAWOLF_API_KEY=${apiKey} npx qawolf test --env '{ "MY_VARIABLE": "secret" }' --env_name 'Staging' --tags 'Sign up,Checkout'`}</CodeBlock>
```
jperl, line 29: I would do --tags, --environment, then --variables. Since I think that is the order of likelihood.

app/components/Docs/CliExample.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <p>
+        You can run your tests with the QA Wolf CLI. Use the command below to
+        run all of your tests:
+      </p>
+      <CodeBlock
+        className="language-bash"
+        style={{ whiteSpace: "break-spaces" }}
+      >{`QAWOLF_API_KEY=${apiKey} npx qawolf test`}</CodeBlock>
+      <p>
+        You can also optionally specify environment variables, an environment
```
jperl, line 23: You can filter the tests to run by tags, choose the environment, and override variables.

app/components/Docs/SuiteApiExample.tsx
```diff
@@ -0,0 +1,21 @@
+import { useContext } from "react";
```
jperl, line 1: Let's cut this from our docs. I am not aware of anyone using it directly.

app/components/Modals/CreateSuite.tsx
jperl: Instead I would just do .catch(closeModal). in handleClick, since someone can delete an environment up to that point too and I'd remove this extra check.

app/components/Modals/Tags/Form.tsx
```diff
@@ -0,0 +1,64 @@
...
+      setNameError(name ? copy.noCommas : copy.required);
+      return;
+    }
+
+    setNameError("");
+
+    if (tag) {
+      updateTag({
+        variables: { id: tag.id, name },
+      }).then(onCancel);
```
jperl, line 41: onClose seems more accurate

app/components/Modals/Tags/TagListItems.tsx
```diff
@@ -0,0 +1,67 @@
...
+  tags,
+  testIds,
+  testTags,
+}: Props): JSX.Element {
+  const [updateTagTests, { loading }] = useUpdateTagTests();
+
+  const handleClick = (tagId: string): void => {
+    if (loading) return;
+
+    const state = getSelectState({ tagId, testIds, testTags });
```
jperl, line 30: Maybe `const isChecked = getSelectState({ tagId, testIds, testTags }) === "all";`

app/components/shared/TriggerIcon.tsx
```diff
@@ -2,28 +2,34 @@ import { Box } from "grommet";
...
+  isApi?: boolean;
+  trigger: ShortTrigger | null;
+};
 
-const getTriggerIconComponent = (trigger: ShortTrigger): Icon => {
-  if (trigger.deployment_provider) return Rocket;
-  if (trigger.repeat_minutes) return Clock;
+const getTriggerIconComponent = (trigger: ShortTrigger | null): Icon => {
+  if (trigger?.deployment_provider) return Rocket;
+  if (trigger?.repeat_minutes) return Clock;
```
jperl, line 17: Is this Plug used anymore? Maybe if neither it should return null?

app/pages/docs/run-tests-with-api.mdx
```diff
@@ -13,56 +11,18 @@ import ViewRunHistory from "../../components/Docs/ViewRunHistory";
...
-- The `trigger_id` in the request body tells QA Wolf to run tests assigned to that trigger: `"trigger_id": "xxxv8y6ii000008mocm506o3l"`
-- The optional `env` object in the request body passes additional environment variables to your tests: `"env": { "MY_VARIABLE": "secret" }`
+- The required <ExternalLink href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization">Authorization header</ExternalLink> provides your team's API key: `-H "Authorization: API_KEY_HERE"`.
+- The optional `tags` string is a comma separated list of test tag names: `"tags": "Checkout,Sign up"`. If not provided, all of your tests will run. Otherwise, tests with any of the provided tags will run.
+- The optional `environment` string in the request body specifies which <Link href="use-environment-variables#create-an-environment">environment</Link> to use when running your tests: `"environment": "Staging"`. If you have multiple environments and do not provide an `env_name`, QA Wolf will error and ask you to specify an environment.
```
jperl, line 24: `environment` not `env_name`?

jperl, line 24: - The optional `environment` string is the name of the <Link href="use-environment-variables#create-an-environment">environment</Link>...

app/components/shared/TriggerIcon.tsx
flaurida: I use it if they trigger tests with API

app/pages/docs/run-tests-with-api.mdx
```diff
@@ -13,56 +11,18 @@ import ViewRunHistory from "../../components/Docs/ViewRunHistory";
...
+- The optional `variables` object in the request body passes additional <Link href="use-environment-variables">environment variables</Link> to your tests: `"variables": { "MY_VARIABLE": "secret" }`.
```
jperl, line 25: passes additional -> provides. 

These will override any environment variables in your --environment.

qawolf/src/cli/testCommand.ts
```diff
@@ -3,14 +3,44 @@ import runTests from "./runTests";
...
+    .option(
+      "-e, --environment <environment>",
+      "environment name to use when running your tests (example: Staging)"
+    )
+    .option(
+      "-v, --variables <variables>",
+      "environment variables to pass to your tests"
+    )
+
+    .option("-s, --no-wait", "do not wait for the tests finish running")
```
jperl: I would remove the `-s`

qawolf/src/cli/index.ts
```diff
@@ -25,3 +25,5 @@ program.allowUnknownOption(false);
 export const runCli = (
   argv: string[] = process.argv
 ): Promise<program.CommanderStatic> => program.parseAsync(argv);
+
+runCli();
```
jperl, line 29: why this?

app/server/api/suites/index.ts
```diff
@@ -2,25 +2,24 @@ import { NextApiRequest, NextApiResponse } from "next";
...
+import { createSuiteForTests } from "../../models/suite";
+import { ensureTeamCanCreateSuite, findTeamForApiKey } from "../../models/team";
+import { findEnabledTestsForTags } from "../../models/test";
+import { findTriggerOrNull } from "../../models/trigger";
+import { CreatedSuite, ModelOptions, Team, Test } from "../../types";
 import { parseVariables } from "../../utils";
 
 // errors example: https://stripe.com/docs/api/errors
-const ensureTriggerAccess = async (
+const findTeamForRequest = async (
```
jperl, line 17: Maybe `ensureTeamForRequest`

app/server/api/suites/index.ts
```diff
@@ -73,6 +55,68 @@ const ensureTriggerAccess = async (
...
+  req: NextApiRequest,
+  options: ModelOptions
+): Promise<CreatedSuite> => {
+  const log = options.logger.prefix("createSuiteForRequest");
+  const team = await findTeamForRequest(req, options);
+
+  const {
+    branch,
+    env,
+    environment: envName,
```
jperl, line 68: I prefer writing these out, environmentName

app/server/api/suites/index.ts
```diff
@@ -73,6 +55,68 @@ const ensureTriggerAccess = async (
...
+    tags,
```
jperl, line 69: should this be tags: tagNames

app/server/api/suites/index.ts
```diff
@@ -73,6 +55,68 @@ const ensureTriggerAccess = async (
...
+    trigger_id,
+    variables,
+  } = req.body;
+  const environment_variables =
+    variables || env ? parseVariables(variables || env) : null;
+
+  let environment_id: string | null = null;
+  let tests: Test[] = [];
+
+  if (trigger_id) {
```
jperl, line 79: I'd leave a comment that `trigger_id` is deprecated, this is just here for backwards compatibility

app/server/models/environment.ts
```diff
@@ -58,6 +63,23 @@ export const deleteEnvironment = async (
   return environment;
 };
 
+export const findDefaultEnvironmentForTeam = async (
```
jperl, line 66: `findOnlyEnvironmentForTeam`?

app/server/resolvers/tag_test.ts
```diff
@@ -0,0 +1,49 @@
...
+    test_ids.map((test_id) =>
+      ensureTestAccess({ teams, test_id }, { db, logger })
+    )
+  );
+  const tagTeam = await ensureTagAccess(
+    { tag_id: add_tag_id || remove_tag_id, teams },
+    { db, logger }
+  );
+
+  // check the test team matches the trigger's team
```
jperl, line 29: matches the tag's team

app/server/models/tag.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  try {
+    ensureValidName(name);
+
+    const teamTags = await findTagsForTeam(team_id, { db, logger });
+
+    const tag = {
+      color: buildColor(teamTags.map((t) => t.color)),
+      id: cuid(),
+      name: name.trimStart(),
```
jperl, line 43: trim()

app/server/models/tag.ts
```diff
@@ -0,0 +1,211 @@
...
+    .whereIn("name", tagNames)
+    .andWhere({ team_id })
+    .orderBy("name", "asc");
+
+  const missingTagName = tagNames.find(
+    (n) => !tags.some((tag) => tag.name === n)
+  );
+  if (missingTagName) {
+    log.error("not found", missingTagName);
+    throw new Error(`Tag ${missingTagName} not found`);
```
jperl, line 112: Should this be a ClientError?

app/server/models/tag.ts
```diff
@@ -0,0 +1,211 @@
...
+};
+
+export const findTagsForTests = async (
+  test_ids: string[],
+  { db, logger }: ModelOptions
+): Promise<TagsForTest[]> => {
+  const log = logger.prefix("findTagsForTests");
+  log.debug("tests", test_ids);
+
+  const tests = await db("tests").select("id").whereIn("id", test_ids);
```
jperl, line 141: I don't think this is necessary

app/server/models/test.ts
```diff
@@ -191,30 +174,73 @@ export const findEnabledTests = async (
...
+    (await findTagIdsForNames({ names: tag_names, team_id }, { db, logger }));
+
+  const tagTestRows = await db("tag_tests")
+    .distinct("test_id")
+    .whereIn("tag_id", finalTagIds);
+  const testIds = tagTestRows.map((row) => row.test_id);
+
+  const tests = await db("tests")
+    .whereIn("id", testIds)
+    .andWhere({ deleted_at: null, "tests.is_enabled": true })
```
jperl, line 200: just is_enabled: true right?

app/server/models/trigger.ts
```diff
@@ -317,6 +292,32 @@ export const getNextAt = ({
     .toISOString();
 };
 
+export const hasTrigger = async (
```
jperl, line 295: hasTriggerOrApiSuite

app/server/models/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  "#C54BDE",
+  "#56BBD6",
+  "#8BC22D",
+  "#E59C59",
+  "#DA4E94",
+  "#ABB3C2",
+  "#667080",
+];
+
+export const buildColor = (
```
jperl, line 12: Inside of teamColors, maybe it's usedColors

app/server/resolvers/tag.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const tagsForTriggerResolver = async (
+  { id, team_id }: Trigger,
+  _: Record<string, unknown>,
+  { db, logger, teams }: Context
+): Promise<Tag[]> => {
+  const log = logger.prefix("tagsForTriggerResolver");
+  log.debug("trigger", id);
+
+  ensureTeamAccess({ logger, team_id, teams });
```
jperl, line 86: Why does this have id and team_id? Also I think you need to ensureTriggerAccess here.

jperl: NVM seems you don't need the ensure access here since it is checked from the parent

jperl (approved):

PR #1288, opened 2021-05-14
Tag tweaks

flaurida:
* fix minor bugs and improve UX


app/components/Dashboard/Tests/index.tsx
```diff
@@ -14,13 +14,17 @@ type Props = {
...
 export default function Tests({ branch, teamId }: Props): JSX.Element {
-  const { filter, tagIds } = useTagQuery();
+  const { filter, tagNames } = useTagQuery();
 
   const [search, setSearch] = useState("");
   const [checkedTestIds, setCheckedTestIds] = useState<string[]>([]);
 
   const { tests: testsData, loading } = useTests({ branch, teamId });
 
+  const testIds = useMemo(() => {
```
jperl, line 24: Why is this in a hook instead of just doing the map?

jperl: Also right below you do the map again

app/components/Modals/Triggers/CreateOrEditTrigger/SelectTags.tsx
```diff
@@ -14,16 +14,13 @@ export default function SelectTags({
   tagIds,
   tags,
 }: Props): JSX.Element {
-  const handleClick = (tagId: string): void => {
+  const handleClick = (tagName: string): void => {
     const newTagIds = [...tagIds];
 
-    const index = newTagIds.indexOf(tagId);
+    const index = newTagIds.indexOf(tagName);
```
jperl, line 20: is this supposed to be names?

app/hooks/mutations.ts
```diff
@@ -655,20 +655,20 @@ export const useDeleteTag = (): MutationTuple<
...
     onCompleted: (response) => {
-      const tagId = response?.deleteTag?.id;
-      if (!tagId || !tagIds.includes(tagId)) return;
+      const tagName = response?.deleteTag?.name;
+      if (!tagName || !tagNames.includes(tagName)) return;
 
-      const newTagIds = [...tagIds];
-      const index = newTagIds.indexOf(tagId);
-      if (index > -1) newTagIds.splice(index, 1);
+      const newtagNames = [...tagNames];
```
jperl, line 667: new**T**agNames

app/server/models/tag_trigger.ts
```diff
@@ -22,16 +22,16 @@ export const updateTagTriggersForTrigger = async (
       })
       .map((tagTrigger) => tagTrigger.id);
 
-    const missingTagIds = tag_ids.filter((tag_id) => {
+    const missingTagNames = tag_ids.filter((tag_id) => {
```
jperl, line 25: Why are these names instead of ids?

app/components/Dashboard/Tests/index.tsx
flaurida: it does infinite render without `useMemo`

PR #1303, opened 2021-05-19
Reduce duplicate clicks; support check/uncheck

### LINKED ISSUE #1302: 🖱️🖱️Grommet check box records two clicks for one
## Bug description

Grommet check box records two clicks for one

## Steps to reproduce

```js
await page.goto('https://v2.grommet.io/checkbox');
```

Click example check box

### LINKED ISSUE #1342: Use check/uncheck for checkbox and radio inputs
## Description

Rather than generating a `click` on checkbox and radio elements, it is more accurate to generate [check](https://playwright.dev/docs/1.10.0/api/class-page/#pagecheckselector-options) and corresponding `uncheck`.

## Proposed solution

Recorder should detect when a click changed the state of a checkbox or radio input, and then replace the preceding click line with check/uncheck as appropriate.

aldeed:
Resolves #1302 
Resolves #1342

Further reading:
- https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events
- https://www.w3.org/TR/pointerevents2/

### DIFF
--- a/.github/workflows/test_recorder.yml
+++ b/.github/workflows/test_recorder.yml
@@ -36,9 +36,13 @@ jobs:
         if: matrix.test == 'lint'
         working-directory: ./recorder
 
+      - run: Xvfb :0 -ac -screen 0 1288x804x24 &
+        if: matrix.test != 'lint'
+
       - name: Test recorder
         run: QAWOLF_BROWSER=${{ matrix.test }} npm test
         if: matrix.test != 'lint'
         working-directory: ./recorder
         env:
           CI: true
+          DISPLAY: ":0.0"
--- a/recorder/package-lock.json
+++ b/recorder/package-lock.json
@@ -27,6 +27,7 @@
         "playwright-core": "1.10.0",
         "prettier": "^2.1.2",
         "rimraf": "^3.0.2",
+        "serve-handler": "^6.1.3",
         "ts-jest": "^26.4.0",
         "ts-loader": "^8.0.4",
         "ts-node": "^9.0.0",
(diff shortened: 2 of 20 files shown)


jperl (approved):
With test 👍

aldeed:
Because the original fix wasn't robust enough, I expanded the scope of this to include the check/uncheck feature work. Here's the current status:

- [x] Avoiding duplicate generation of `page.click` for a checkbox or radio
- [x] Generating `page.check` or `page.uncheck` to replace the `page.click`
- [ ] Running the generated `page.uncheck` doesn't always work. See Playwright bug: https://github.com/microsoft/playwright/issues/6850
- [ ] More manual testing needed on: fancy radio (where actual input is hidden), vanilla radio, vanilla checkbox
- [ ] Update tests

aldeed:
@jperl Can you do another review pass since this has increased in scope and complexity? check/uncheck is now implemented, as well as solving duplicate click on all browsers.

- Clicking input of type checkbox or radio will produce only one `click` action
- IF the input is visible (according to playwright), we also produce a `check` or `uncheck` action. We shouldn't use this action if it isn't visible because it will fail when running later.
- New patch logic in the runner will replace the previous `click` line with the `check`/`uncheck` line, if it can determine that they were the same source event. It usually can, and it usually happens fast enough that you never see the `click` line.

The upshot is this:
- for Grommet checkbox, you'll see one `click` line added
- for MUI checkbox and HTML checkbox, you'll see one `check` or `uncheck` line added
- I didn't test other component libs, but it should be either of the above outcomes, depending on how they are handling the underlying `input` element.

PR #1305, opened 2021-05-20
Render take 2

### LINKED ISSUE #1306: Trigger tests from Render
Add [Render](https://render.com/) trigger for our users that have stacks that cannot run on Netlify / Vercel, but still want to run e2e tests against PR previews.

flaurida:
Resolves #1306


app/components/Dashboard/Tests/index.tsx
```diff
@@ -25,7 +25,9 @@ export default function Tests({ branch, teamId }: Props): JSX.Element {
   const { data: tagsData, loading: tagsLoading } = useTagsForTests({
     test_ids: testIds,
   });
-  const testTags = tagsData?.tagsForTests || [];
+  const testTags = useMemo(() => {
```
jperl, line 28: Did you intent to make this change? If so can you leave a comment as to why useMemo is needed.

app/components/Docs/GitHubComment.tsx
```diff
@@ -0,0 +1,20 @@
+import Image from "./Image";
+
+export default function GitHubComment(): JSX.Element {
+  return (
+    <>
+      <p>
+        QA Wolf will also comment on a pull request with the most recent test
```
jperl, line 7: ~also~

app/server/api/github/deployment.ts
```diff
@@ -1,48 +1,79 @@
...
 };
 
+export const buildUrl = ({
+  deploymentUrl,
+  pullRequestId,
+  trigger,
+}: BuildUrl): string => {
+  if (!pullRequestId || !trigger.deployment_preview_url) return deploymentUrl;
+
+  return trigger.deployment_preview_url.replace(
```
jperl, line 59: I'd probably use URL to grab the host then parse from there, in case they do not provide the protocol.

const url = new URL(trigger.deployment_preview_url);
const prHost = url.host.replace(/\d+.onrender.com\/?/, `${pullRequestId}.onrender.com`);
return `https://${prHost}`;

jperl (approved):

Issue #1271, opened 2021-04-29
Snippet for asserting a value


flaurida:
Allow asserting that an element's value is as expected - ie `await page.$eval('#my-id', element => element.value); assert(value === "my value");`

PR #1319, opened 2021-05-21
Allow text selectors in assertText

flaurida:
#1271
#1321


app/components/Editor/contexts/TestContext.tsx
```diff
@@ -73,14 +73,14 @@ export const TestProvider: FC = ({ children }) => {
 
   // tee up correct branch and environment if test edited
   useEffect(() => {
-    if (suite?.environment_id && suite.environment_id !== environmentId) {
+    if (suite?.environment_id) {
```
flaurida, line 76: fixes bug where you can't change environment after suite loads

Issue #1287, opened 2021-05-13
👥 Support multiplayer


jperl:
**Problem**

Users need to be able to collaborate on tests and helpers at the same time. Right now helpers will get overridden, additionally if a runner is not attached test code will get overwritten.

**Solution**

Support google-docs like collaborative text editing always.

- [ ] Add e2e for newly created test code gen
- [ ] Add e2e test for multiplayer across context

PR #1326, opened 2021-05-24
Revamp editor endpoints

flaurida:
Relates to #1287


app/server/models/test.ts
```diff
@@ -373,7 +373,7 @@ export const updateTest = async (
     }
 
     const updates: Partial<Test> = {
-      code: code || existingTest.code,
+      code: isNil(code) ? existingTest.code : code,
```
jperl, line 376: Seems like it should not set it if the code is not passed

PR #1327, opened 2021-05-24
Restructure editor context and models

jperl:
Relates to #1287

- Make test and helpers into a common FileModel that we can swap to new state management this week

- Hook up to new file query and mutation, run query, and commitChanges

- Clean up the contexts. Consolidate TestContext and EditorController into an EditorContext, RunContext into run and suite hooks

- Remove mode hook, use EditorContext runId to determine if it is a run or not.

Some fixes to save logic for the meantime until we change out the implementation this week:

- Propagate all changes regardless of sender (might fix code gen not being saved until user change)

- Wait to connect change / auto-save logic until after file is loaded (might fix a bug where changes can be overwritten when a new user joins)


app/components/Editor/Sidebar/HelpersEditor.tsx
```diff
@@ -16,17 +14,10 @@ export default function HelpersEditor({
...
-  const { controller, hasWriteAccess } = useContext(TestContext);
+  const { helpersModel, isHelpersReadOnly } = useContext(EditorContext);
 
-  const [monaco, setMonaco] = useState<typeof monacoEditor | null>(null);
-
-  useEnvTypes({ env, monaco });
-
-  const editorDidMount = ({ editor, monaco }) => {
-    controller.setHelpersEditor(editor);
-    setMonaco(monaco);
```
jperl, line 28: setMonaco and includeTypes are unnecessary since they are set globally in CodeEditor and it is constructed always now (but hidden)

app/lib/types.ts
```diff
@@ -325,6 +325,11 @@ export type RunProgress = {
   status: RunStatus;
 };
 
+export type Selection = {
+  startLine: number;
```
flaurida, line 329: abc

flaurida (approved):

PR #1335, opened 2021-05-28
Multiplayer - client updates

jperl:


runner/src/code/FileModel.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const debug = Debug("qawolf:FileModel");
+
+type ConnectOptions = {
+  authorization: string;
+  id: string;
+  url: string;
+};
+
+export class FileModel extends EventEmitter {
```
jperl, line 16: remove event emitter

PR #1337, opened 2021-05-28
Multiplayer API changes

flaurida:


jperl (approved):
We'll probably want to disambiguate between a github file and our file model, but we can do that later :)

PR #1338, opened 2021-05-30
Delete file resolver

flaurida:


app/server/resolvers/file.ts
```diff
@@ -41,6 +41,30 @@ type FormatTestFile = {
...
+  }
+
+  if (type === "runhelpers") {
+    await ensureSuiteAccess({ suite_id: id, teams }, { db, logger });
+  }
+
+  if (type === "test") {
+    await ensureTestAccess({ teams, test_id: id }, { db, logger });
+  }
+};
```
jperl, line 66: throw an error at the end

PR #1346, opened 2021-06-11
Prefer input name attribute over type attribute

### LINKED ISSUE #1331: Do not prefer [type="text"] over [name="email"] in text inputs
also for `[type="email]` and `[type="tel"]`

aldeed:
Fixes #1331 

Primary fix: When choosing a selector for an input, the `name` attribute will be preferred over the `type` attribute.

Secondary fix: To ensure `name` is used for common fields like `address1` or `phone1`, a valid word followed by a single number is no longer considered to be dynamic, and thus is not ignored as a possible good selector.

### DIFF
--- a/recorder/src/getCues.ts
+++ b/recorder/src/getCues.ts
@@ -37,7 +37,8 @@ const PENALTY_MAP = {
   tagnth: 31,
   text: 10,
   title: 10,
-  type: 10,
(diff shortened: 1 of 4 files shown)


aldeed:
@jperl This is targeting my other PR for purposes of the diff, but the commits should be easy to cherry pick into develop.

PR #1349, opened 2021-06-11
Support element choosing on multiple tabs and iframes

### LINKED ISSUE #1347: Choosing an element on second tab still uses first tab's page variable
## Bug description

Choosing an element on second tab still uses first tab's page variable

## Steps to reproduce

Load any website. Click a link that opens in a new tab. Activate the element chooser and then choose any element on the new tab. Note that `page` is used as the action variable rather than `page2`.

aldeed:
Resolves #1347 

- Choosing an element on any tab will generate code that uses the correct page variable to target that tab
- Choosing an element in any iframe will generate code that uses the correct frame variable to target that frame

Technical details:
- give ElementChooser instance a reference to `this._variables`
- when emitting element chooser events, search `this._variables` to find the correct variable name string based on incoming `frame` and `page` instances from Playwright
- in the app component, use `event.variable` instead of hard coding `page` (fall back to old behavior)
- also pass any `initializeCode` to client so that it can insert it with the snippet

### DIFF
--- a/app/components/Editor/Sidebar/Snippet/Action.tsx
+++ b/app/components/Editor/Sidebar/Snippet/Action.tsx
@@ -2,8 +2,9 @@ import { Box } from "grommet";
 
 import { copy } from "../../../../theme/copy";
 import Text from "../../../shared/Text";
-import { ActionType, buildActionOptions, labelProps } from "./helpers";
+import { ActionType, buildActionOptions } from "./helpers";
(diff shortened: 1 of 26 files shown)


aldeed:
Seems to work for tabs but I haven't tested iframes or added Jest tests yet.

aldeed:
Manual testing is looking good. All that's left is to add a few more tests.

PR #1360, opened 2021-10-03
simplify project

jperl:
Simplify the project to the open source library our customers can use independently to reduce confusion


robin-rpr:
Simplifying project is a fun way of turning open source into closed-source than _"reducing confusion"_

PR #1390, opened 2026-07-09
fix(runner): validate outer hop `node_modules` against declared deps

### LINKED ISSUE #123: ✅Add checkbox & 🔘radio support

chajac:
>[!NOTE]
PR body AI drafted & edited as needed
# Overview of Changes

`qawolf flows run` could fail with `ERR_MODULE_NOT_FOUND` for a dependency the flow project correctly declares. Root cause: the outer hop discovery in `populateOuterHop` walked up from the flow project and symlinked the **first** ancestor `node_modules` it found, with no check that it actually contains the project's deps, and finding one suppressed the `npm install` fallback that would have worked. 

## How it works

- **Validate then walk** (`outerHop.ts`): a candidate `node_modules` is symlinked only if every installable declared dep resolves in it (pinned executor packages are excluded, the inner hop owns those). Failing candidates are recorded and the walk continues; if none satisfies, the existing fallback installs the project's declared deps into the run dir (executor packages stripped, so `workspace:*` pins don't break the install).
- **Observability**: `populateOuterHop` returns a result union (`symlink | install | none`) threaded through `prepareRunDir` > `flows run`, which renders `Installing N project dependencies…` before the (potentially slow) fallback install and debug logs each rejected candidate with the deps it was missing.
- **Actionable failure**: residual module resolution failures get a hint naming the unresolved package and pointing at the project's `package.json`.
Path like specifiers (broken `./relative` imports) are excluded so the hint never suggests "installing" a file path.

# Testing

- New unit & integration tests cover nested bundle, hoisted monorepo, split hoisting, sibling repo, and fallback install topologies.
- A/B verified on real builds: the prefix CLI reproduces the reported error on all three channels (node, bun, compiled binary). The fixed CLI passes, including against a real pulled bundle with 15 deps.

| Topology | Pre-fix | Fixed |
  |---|---|---|
  | Flow project nested in another repo | ❌ `Cannot find package 'date-fns'` (3/3 channels) | ✅ pass — `Installing 1 project dependency…` (3/3) |
  | Monorepo, deps hoisted to workspace root | ✅ pass | ✅ pass (no regression) |
  | Standalone project, own `node_modules` | ✅ pass | ✅ pass (no regression) |
  | Real pulled bundle (15 deps, `latest` + `workspace:*` specs), nested | ❌ `ERR_MODULE_NOT_FOUND: Cannot find package 'date-fns'` | ✅ pass — `Installing 15 project dependencies…` |

# Checklist

- [x] Self-review completed

### DIFF
--- a/.changeset/validate-outer-hop-discovery.md
+++ b/.changeset/validate-outer-hop-discovery.md
@@ -1,0 +1,8 @@
+---
+"@qawolf/cli": patch
+---
+
+Fix `ERR_MODULE_NOT_FOUND` for correctly declared flow dependencies: dependency
+discovery now validates an ancestor `node_modules` before reusing it, and falls
+back to installing the project's declared deps when none satisfies them
+(reported as `Installing N project dependencies…`).
--- a/.changeset/validate-outer-hop-discovery.md
+++ b/.changeset/validate-outer-hop-discovery.md
@@ -2,7 +2,4 @@
 "@qawolf/cli": patch
 ---
 
-Fix `ERR_MODULE_NOT_FOUND` for correctly declared flow dependencies: dependency
-discovery now validates an ancestor `node_modules` before reusing it, and falls
-back to installing the project's declared deps when none satisfies them
-(reported as `Installing N project dependencies…`).
+Fix `ERR_MODULE_NOT_FOUND` for correctly declared flow dependencies: dependency discovery now validates an ancestor `node_modules` before reusing it, and falls back to installing the project's declared deps when none satisfies them (reported as `Installing N project dependencies…`).
--- a/src/domains/runtimeEnv/outerHop.test.ts
+++ b/src/domains/runtimeEnv/outerHop.test.ts
@@ -252,6 +252,37 @@ describe("populateOuterHop", () => {
     expect(result).toEqual({ mode: "none" });
   });
 
+  it("never consults a sibling directory's node_modules (multi-repo layout)", async () => {
+    const root = await makeTmpDir();
+    // The sibling repo's node_modules WOULD satisfy the project's deps — but
+    // it is not an ancestor of the project, so discovery must never see it.
+    const sibling = join(root, "sibling-repo");
+    await seedNodeModules(sibling, ["date-fns"]);
+    const projectDir = join(root, "flows-repo");
+    await writePackageJson(projectDir, { "date-fns": "2.29.3" });
+    const runDir = join(root, "run");
+    await mkdir(runDir, { recursive: true });
+
+    installMock.mockImplementation(async () => ({ exitCode: 0, stderr: "" }));
+
+    const result = await populateOuterHop({
+      projectDir,
+      runDir,
+      fs: makeDefaultFs(),
+      install: installMock,
+    });
+
+    // No ancestor satisfies, so the project's own deps are installed — the
+    // satisfying sibling tree is invisible to the upward walk.
+    expect(result.mode).toBe("install");
+    if (result.mode === "install") {
+      expect(result.rejected.map((r) => r.dir)).not.toContain(
+        join(sibling, "node_modules"),
+      );
+    }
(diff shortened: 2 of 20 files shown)


src/domains/runtimeEnv/outerHop.test.ts
```diff
@@ -0,0 +1,306 @@
...
+): Promise<void> {
+  await mkdir(dir, { recursive: true });
+  await writeFile(
+    join(dir, "package.json"),
+    JSON.stringify({ name: "flows-project", dependencies }),
+  );
+}
+
+const installMock =
+  mock<(cwd: string) => Promise<{ exitCode: number; stderr: string }>>();
```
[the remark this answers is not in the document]
chajac: [7fe31e1](https://github.com/qawolf/cli/pull/1390/commits/7fe31e14f04f0071771037370c809db81eca7338)

.changeset/validate-outer-hop-discovery.md
```diff
@@ -0,0 +1,8 @@
+---
+"@qawolf/cli": patch
+---
+
+Fix `ERR_MODULE_NOT_FOUND` for correctly declared flow dependencies: dependency
+discovery now validates an ancestor `node_modules` before reusing it, and falls
+back to installing the project's declared deps when none satisfies them
+(reported as `Installing N project dependencies…`).
```
michael-pr: We can remove the hard-wrapping here (and in PR description).

src/domains/runtimeEnv/outerHop.test.ts
```diff
@@ -0,0 +1,309 @@
+// oxlint-disable eslint/max-lines -- test file covers all candidate-validation scenarios; splitting would fragment coverage
```
michael-pr: Let's see if we can avoid this by moving some of the test setup into a shared helper?

src/domains/runtimeEnv/outerHop.ts
```diff
@@ -75,7 +123,12 @@ async function readInstallableDeps(
     return {};
   }
 
-  const parsed: unknown = JSON.parse(content);
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(content);
+  } catch {
+    return {};
```
michael-pr, line 132: We should throw here instead or add a single line comment explaining why.

src/domains/runtimeEnv/prepareRunDir.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable eslint/max-lines -- test file for all prepareRunDir scenarios including nested-bundle discovery in Task 3 */
```
michael-pr: Nit: similar comment here, let's see what we can do to be under this rule.

michael-pr (approved):
Mostly suggestions, nothing blocking. Good to merge after addressing them.

src/domains/runtimeEnv/prepareRunDir.test.ts
chajac: My oversight on those, fixed: [604725d](https://github.com/qawolf/cli/pull/1390/commits/604725db6b38ab1a6322471413eb167b25bd6471)

src/domains/runtimeEnv/outerHop.test.ts
chajac: [604725d]

src/domains/runtimeEnv/outerHop.ts
chajac: Opting for a clear comment:
[2dc7920](https://github.com/qawolf/cli/pull/1390/commits/2dc79202297e0d4c43ff458a74ed2a68d5c0d62d)

.changeset/validate-outer-hop-discovery.md
chajac: Done & done [812ee56](https://github.com/qawolf/cli/pull/1390/commits/812ee56bfd70409790b5ee5de3633196a4229fa7)

PR #1421, opened 2026-07-30
refactor(cli): give the win32 shim-name policy one home

smonn:
Resolves WIZ-11296

# Overview of Changes

`appiumBins.ts` and `playwrightBins.ts` held the same 13 lines with one string changed, including a byte-identical comment. Both collapse into `core/nodeModulesBins.ts`, which owns `nodeModulesBinCandidates(envDir, name, platform)`; `appiumCliCandidates` and `playwrightCliCandidates` stay as wrappers, so the six consumers changed only their import path.

The motivation is `src/shell/spawn.ts:45`, which tests the resolved name with `/\.(cmd|bat)$/i`. Every producer of a shim name has to stay in sync with that regex by hand, and one producer is easier to keep in sync than two identical copies. `androidBins.ts` stays separate — a different suffix rule over different directories, not `node_modules/.bin`.

# Testing

Structural change only, so the merged test file is the safety net: it asserts the shared helper directly, then pins each wrapper's tool name in both the POSIX and win32 branches against literal expectations.

```bash
bun run typecheck
bun run lint
bun run format:check
bun run knip
bun run test
bun run build
```

# Checklist

- [x] Self-review completed


smonn:
@coderabbitai review

chajac (approved):

PR #1445, opened 2026-08-05
feat(runner): see, act on and keep alive a runner from the terminal

gorangajic:
Relates to NOVA-1365
Stacked on #1444 — review that first.

# Overview of Changes

The first runner PR could start a runner and run flows on it, but a caller could not see the browser or touch it. Nothing could take a screenshot, click, type, or evaluate an expression against the live page, and a runner would be reaped while its caller was still thinking.

This completes the group with `screenshot`, `act`, `exec` and `keepalive`. A screenshot is decoded and written to disk as a real image, so any harness can open it. `act` performs one action per call in the computer-use vocabulary a vision model already emits — underscores and all — so an agent forwards its model's tool call instead of translating it into flags, and a whole action can be piped in as JSON. `exec` evaluates a snippet against the live page. `keepalive` resets the inactivity clock with a cheap read, and never launches a runner: someone asking to keep one alive already has one in mind.

A screen that is still coming up and a runner that has no screen at all read differently and exit differently — one is worth retrying, the other never will be.

# Blocked on an unpublished contracts version

**This will not pass CI yet, and that is expected.** The commands call `runner.*` contracts that no published `@qawolf/api-contracts` version carries — they exist only in unmerged platform PRs. The pin stays at main’s `0.20.0` rather than moving to a version that does not exist; once the platform publishes the first version carrying the `runner.*` contracts, a one-line bump turns CI green.

Until then `bun run generate` fails on a missing export, which takes `typecheck`, `lint`, `test` and `build` down with it. `format:check` and `knip` pass. Every gate passes locally against a contracts build staged with `bun run stage:api-contracts`.

# Testing

```bash
bun run stage:api-contracts ../platform   # a checkout carrying the runner contracts
bun run typecheck
bun run lint
bun run format:check
bun run knip
bun run test
bun run build
```

# Checklist

- [x] Self-review completed


MateusZitelli (approved):
🔥

Dhaaaf (approved):

PR #1537, opened 2026-08-29
feat(cli): add qawolf flows lint to check flow files against QA Wolf's rules

JAllen2022:
# Overview of Changes

`qawolf flows lint [pattern]` lints source files with QA Wolf's own rules - the same linter the platform editor and Tester run, imported from `@qawolf/workflow-linter`. It sits in the `flows` group beside `list`, `pull`, and `run`, and expands patterns through the same `expandPatterns` helper (`node_modules` excluded, `.qawolf/<env>/` caches included). No pattern means every `.ts`/`.js` file in the project - flows, helpers, and page objects alike, since the POM rules a team's `.eslintrc.json` enables target exactly the non-flow files. The lint domain owns lintability: patterns select candidates, and `selectLintableFiles` keeps `.ts`/`.js` (same extension test as the platform's `lintFilePath`), so a pattern matching a `.json` never gets parsed as TypeScript. It reads the project's `.eslintrc.json` (severity-only overrides, bundled plugins only, same contract as the editor) by walking up from cwd, bounded by the package root via `resolveProjectDirSafe`. Output is eslint-stylish per file plus a problem count. Type-aware rules work - the command builds a real TS program over each file's import graph, so things like `no-unnecessary-type-assertion` fire across files.

Exit codes: 0 for clean or warnings-only, `testFailure` (1) when any error-level finding exists, `invalidArgs` (2) when a pattern matches no lintable file - including a pattern that matches only non-lintable files, which would otherwise pass a gate it never checked. **Zero matches follows `flows run`, not `flows list`** - the two siblings disagree, and lint is a CI gate like run: a typo'd pattern silently passing is the failure worth catching. `--allow-no-match` downgrades it to 0, same flag as run. Mixed matches lint the lintable subset silently, like eslint.

Registered as a `local` command (no auth, no API), so the generated skill table advertises it to agents as safe to run and retry. Minor changeset included.

# Testing

```bash
bun run typecheck
bun run lint
bun run format:check
bun run knip
bun test
bun run generate
```

1874 tests pass, 18 of them lint's. Also verified end to end with the built bundle (`node dist/cli.js`) in a temp project: a bare run reports the flows, a helper, and a `src/pages/` page object while `.json`/`.md` files never appear; a pattern narrows; a `.json`-only pattern is a no-match (exit 2 with the `--allow-no-match` hint, 0 with the flag); `.eslintrc.json` downgrading a rule to warn turns all findings into warnings + exit 0; cross-file type-aware findings fire; and a parent directory's `.eslintrc.json` outside the project is ignored.

# Checklist

- [x] Self-review completed


skills/qawolf-cli/SKILL.md
```diff
@@ -140,6 +140,7 @@ that `url`; never guess a route and never send a repository link in its place.
 | `qawolf environment update` | write | Update an environment owned by the caller's team and return it in the environment.get shape. Omitted fields remain unchanged. |
 | `qawolf flow addTag` | write | Assign an existing tag to the selected flows. Create tags with tag.create. |
 | `qawolf flow update` | write | Move a flow between draft and active readiness. The other statuses shown in the app are derived and cannot be set. |
+| `qawolf flows lint` | local | Lint source files matching [pattern], or every .ts/.js file when omitted, with QA Wolf's rules, honoring the project's .eslintrc.json |
```
gorangajic, line 143: Does this belong on the `flows` namespace?

gorangajic: I am not even sure if we need this, if we publish our own rules and add them to the `.eslintrc.json` as a package, users can use eslint command directly instead of going throught our cli

PR #1550, opened 2026-09-02
fix(runner): warn when a freshly launched runner is shadowed by QAWOLF_RUNNER_ID (ARC-661)

erzhtor:
## Overview of Problem

`qawolf runner launch` remembers the runner it just started as this directory's `.qawolf`-stored default, but `QAWOLF_RUNNER_ID` outranks that default when resolving a runner-less command. A stale `QAWOLF_RUNNER_ID` — inherited from a harness, or exported earlier in the same shell for a different runner — silently keeps every later command targeting the old runner instead. Surfaced during mobile dogfooding: launching an Android runner while `QAWOLF_RUNNER_ID` still named the session's default Playwright runner cost several debug cycles before the cause was found, since nothing errored.

## Overview of Changes

- `runner launch`: warn on stderr when it succeeds but `QAWOLF_RUNNER_ID` names a different id than the one just launched, naming both ids and the fix (`--runner <new>` or re-export `QAWOLF_RUNNER_ID=<new>`)
- No change to runner resolution order — considered making the directory default outrank env instead, but that would regress Tester's multi-runner "slot" feature (every `runner launch` for an additional slot overwrites the directory default; Tester relies on env staying authoritative to route back to its main runner)

### Verify it works

Added 2 unit tests covering: warns when `QAWOLF_RUNNER_ID` differs from the launched id; does not warn when it already matches or is unset. `bun run test src/domains/interactiveRunner` — 234/234 pass.

## Release Notes

None — CLI-only stderr message, no behavior change.

### DIFF
--- a/src/core/messages/interactiveRunner/lifecycle.ts
+++ b/src/core/messages/interactiveRunner/lifecycle.ts
@@ -7,6 +7,8 @@ export const lifecycleMessages = {
     `Runner ${id} could not be written to .qawolf as this directory's default, so later commands will not find it on their own. Pass --runner ${id}, or set QAWOLF_RUNNER_ID=${id}.`,
   defaultNotForgotten: (id: string) =>
     `Runner ${id} was terminated, but it could not be removed from .qawolf as this directory's default, so later commands will still be sent to it. Pass --runner, or launch a new runner.`,
+  envRunnerIdShadowsLaunch: (launchedId: string, envId: string) =>
+    `QAWOLF_RUNNER_ID=${envId} is still set, so commands that omit --runner will keep targeting it instead of the runner just launched. Pass --runner ${launchedId}, or export QAWOLF_RUNNER_ID=${launchedId}.`,
   keptAlive: (id: string) =>
     `Runner ${id} is alive, and its inactivity clock has been reset.`,
   launchFailed: (id: string, error: string) =>
--- a/src/domains/interactiveRunner/launch.ts
+++ b/src/domains/interactiveRunner/launch.ts
@@ -8,6 +8,7 @@ import { failureFields } from "~/shell/platform/requestWithRetry.js";
 
 import type { InteractiveRunnerDeps } from "./deps.js";
 import { launchAndRemember } from "./launchAndRemember.js";
+import { runnerIdEnvironmentVariable } from "./resolveRunner.js";
 import { parseRunnerId, parseRunnerName } from "./runnerIds.js";
 
 export async function handleRunnerLaunch(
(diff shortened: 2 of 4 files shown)


src/core/messages/interactiveRunner/lifecycle.ts
```diff
@@ -7,6 +7,8 @@ export const lifecycleMessages = {
     `Runner ${id} could not be written to .qawolf as this directory's default, so later commands will not find it on their own. Pass --runner ${id}, or set QAWOLF_RUNNER_ID=${id}.`,
   defaultNotForgotten: (id: string) =>
     `Runner ${id} was terminated, but it could not be removed from .qawolf as this directory's default, so later commands will still be sent to it. Pass --runner, or launch a new runner.`,
+  envRunnerIdShadowsLaunch: (launchedId: string, envId: string) =>
+    `QAWOLF_RUNNER_ID=${envId} is still set, so commands that omit --runner will keep targeting it instead of the runner just launched. Pass --runner ${launchedId}, or export QAWOLF_RUNNER_ID=${launchedId}.`,
```
theonly1me, line 11: Tester's prompt tells the agent a second runner is `$QAWOLF_RUNNER_ID-<name>` and to start it with `qawolf runner launch --id "$QAWOLF_RUNNER_ID-checkout"`, so this fires on every slot launch, and the `export` half is advice we don't want followed there. I think we should drop that half and keep `--runner`.

```suggestion
    `QAWOLF_RUNNER_ID=${envId} is still set, so commands that omit --runner will keep targeting it instead of the runner just launched. Pass --runner ${launchedId} to address this one.`,
```

src/domains/interactiveRunner/launch.ts
```diff
@@ -37,6 +38,20 @@ export async function handleRunnerLaunch(
     return { ...failureFields(launched), exitCode: launched.exitCode };
   }
 
+  // The directory default this launch just wrote is only what a runner-less
+  // command reaches for; QAWOLF_RUNNER_ID outranks it (see resolveRunner.ts),
+  // so a stale value there would silently keep sending those commands
+  // elsewhere. Say so now, while it's this launch's id that's fresh in mind.
+  const envRunnerId = deps.env[runnerIdEnvironmentVariable]?.trim();
```
theonly1me, line 45: `skills/qawolf-cli/references/runner.md` already documents this trap in the resolution order section, can we add a line there that a launch whose id differs from `QAWOLF_RUNNER_ID` now says so on stderr? Tester gets `SKILL.md` handed to it at pod start, so it'll meet this warning with nothing in the docs explaining it.

gorangajic (approved):

src/core/messages/interactiveRunner/lifecycle.ts
erzhtor: Good point, addressed in follow-up here https://github.com/qawolf/cli/pull/1552.

src/domains/interactiveRunner/launch.ts
erzhtor: https://github.com/qawolf/cli/pull/1550#discussion_r3914626652

PR #1578, opened 2026-09-07
feat(runner): write a screenshot to stdout with --out -

gorangajic:
Relates to WIZ-11974.

# Overview of Changes

`qawolf runner screenshot` only wrote a file. The Tester session, which wants the bytes in a process, had to reserve a temp path, hand it to the sandboxed user, run the command, read the file back and delete it on every computer-use step.

`--out -` now writes the decoded JPEG to stdout, the same `-` that already means stdin for `act` and `exec`. Stdout carries the image alone: the confirmation, and the JSON line under `--json`, goes to stderr so nothing follows the bytes into a reader that treats stdout as the file. A terminal on stdout is refused, since nothing there can read the bytes. The help text, the runner guide and the skill say so.

Raw bytes rather than a base64 field in the JSON answer, because the repo's rule is to decode next to the write so nobody files base64 text as a JPEG, and because a byte stream is what a reader that pipes the command into a file or a viewer expects.

# Testing

```bash
bun run typecheck
bun run lint
bun run format:check
bun run knip
bun run test
bun run build
```

Added tests: the writer sends decoded bytes to stdout and nothing to disk, refuses a non-JPEG answer for stdout too, and reports a closed pipe without letting the stream's error event crash the process (the doubles are real Writable streams for that reason); the handler keeps every confirmation off stdout in json and agent mode, refuses a terminal on stdout, and names stdout rather than a file when the pipe fails.

# Checklist

- [x] Self-review completed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/domains/interactiveRunner/takeScreenshot.stdout.test.ts
+++ b/src/domains/interactiveRunner/takeScreenshot.stdout.test.ts
@@ -1,0 +1,101 @@
+import { describe, expect, it } from "bun:test";
+
+import { handleRunnerScreenshot } from "./takeScreenshot.js";
+import { makeAuthCtx, makeTestDeps } from "./deps.testUtils.js";
+
+const jpegBytes = Uint8Array.from([0xff, 0xd8, 0xff, 0xe0, 0x00, 0x10]);
+const imageJpegBase64 = Buffer.from(jpegBytes).toString("base64");
+
+describe("handleRunnerScreenshot --out -", () => {
+  // The session reads the bytes off the pipe instead of reserving a
+  // file, handing it to the sandboxed user, reading it back and deleting it.
+  it("writes the decoded image bytes to stdout and no file", async () => {
+    const { callPublicApi, ctx } = makeAuthCtx();
+    callPublicApi.mockResolvedValue({
+      ok: true,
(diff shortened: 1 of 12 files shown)


src/domains/interactiveRunner/takeScreenshot.stdout.test.ts
```diff
@@ -0,0 +1,119 @@
+import { describe, expect, it } from "bun:test";
+
+import { handleRunnerScreenshot } from "./takeScreenshot.js";
+import { makeAuthCtx, makeTestDeps } from "./deps.testUtils.js";
+
+const jpegBytes = Uint8Array.from([0xff, 0xd8, 0xff, 0xe0, 0x00, 0x10]);
+const imageJpegBase64 = Buffer.from(jpegBytes).toString("base64");
+
+describe("handleRunnerScreenshot --out -", () => {
+  // The Tester session reads the bytes off the pipe instead of reserving a
```
theonly1me: `Tester` is an internal jargon, since CLI is a public repo, it would be good not to expose it:


```suggestion
  // The session reads the bytes off the pipe instead of reserving a
```

theonly1me (approved):

PR #1598, opened 2026-09-11
feat(agent): add --follow to agent send and get

chajac:
> [!NOTE]
> PR body AI drafted & edited as needed
> Stacked on #1597. Lots of this PR is test additions (it is still a large PR though)

# Overview of Changes

`agent.get` reports only that a session is working, and a session that asks a question cannot be answered from the terminal. This PR replaces the generated `agent send` and `agent get` with hand-written commands that take `--follow`. A follow stays attached, prints each reply once as it arrives, prompts for an answer when the AI asks, and exits when the session settles.

**`src/commands/agent/*.ts`**: Commander wiring for `agent send <message>` and `agent get [session]`. `--environment-id` reads `QAWOLF_ENVIRONMENT` and `--workspace-id` is available on both, as the generated commands had them.

**`src/domains/agent/followSession.ts`**: The poll loop. It prints only replies past what it has printed, shows a wait line between them, and stops on a settled status. The timeout is elapsed time less the time spent at a prompt. Ctrl-C stops the follow and says how to pick the session back up.

**`src/domains/agent/answerQuestion.ts`**: The question step. In a terminal it prompts with a list or free text and sends the answer to the same session, with the workspace. Elsewhere it prints the question and the command that answers it, and exits 0.

**`src/domains/agent/send.ts`**, **`src/domains/agent/get.ts`**: The handlers. `send` validates the message through the contract schema before any request. When it continues a session and follows it, it reads the reply count first, so the follow does not re-ask the question that message answered and still prompts for one that arrives later. `get --follow` follows from the read it already made. Both keep the exit code the platform client returns.

**`--file-paths <path...>`** on `agent send`: names files uploaded with `qawolf file requestUpload`, which contracts 0.53.0 added to `agent.send`. Verified live: a CSV of three journeys uploaded to team storage, sent with the flag, and the AI's first reply listed the three journeys from the file.

**Follow-up, not in this PR:** `agent.get` in 0.53.0 also takes a `cursor` and returns `nextCursor`. The follow still reads the whole transcript each poll, which the contract allows when the cursor is omitted. Adopting the cursor, and the hold-until-final rule for a reply that grows in place, is deferred until a real session shows partial replies.

**`src/domains/agent/settlement.ts`**: In `--json` and `--agent` mode a follow ends with the whole session object on stdout, the same object a plain `agent get` answers with.

**`src/shell/agent/sessionStore.ts`**: `agent send` remembers its session in `.qawolf/agent.json`, so a later `agent get --follow` needs no id.

**`src/domains/publicApi/skippedContracts.ts`**: `agent.get` and `agent.send` leave the generator, with the reason recorded.

**`src/commands/qawolfCliSkill.template.md`**: Agent-facing guidance: pass `--follow`, do not poll, expect quiet stretches, read the final status from the closing JSON line, and use a file for a long message.

**`.changeset/agent-follow.md`**: Declares the two flag renames from the generated commands: the message is now the first argument, and `--session-id` is now `--session`.

# Testing

```bash
bun run typecheck
bun run lint
bun run knip
bun run test
bun run build:binary
```

- New tests cover the follow loop, the re-prompt guard, the pre-send reply count, the wait line ordering, the elapsed-time timeout, machine-mode output, exit code passthrough, and the workspace on answers.
- The full suite passes: 2456 tests.
- Verified against a mock platform in a pty: the terminal flow with a choice prompt, a piped run that ends with the session object, and a Ctrl-C during the wait that prints the detach line and exits 130.
- Verified against the real platform: a session started, asked a question, took the answer from the terminal, and continued.

# Checklist

- [x] Code follows the style in AGENTS.md
- [x] Self-review done
- [x] Tests added or updated
- [x] No breaking changes


src/commands/qawolfCliSkill.template.md
```diff
@@ -129,6 +129,54 @@ changes team state; `local` only affects this machine. A parenthesized
...
+exits when the session settles: 0 when the work is complete, non-zero when it
+failed or was cancelled. Looping `qawolf agent get` yourself costs a round trip
+per tick and shows you replies you have already seen.
+
+A session can stop and ask a question. With `--follow` in `--agent` or `--json`
+mode the CLI prints the question and **exits 0** — a session that asked something
+has handed the work back, it has not failed. Answer it, then pick the session
+back up:
+
+```
```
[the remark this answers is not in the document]
chajac: Fixed in 143a08356.

src/domains/agent/send.ts
```diff
@@ -0,0 +1,115 @@
...
+    publicContractsV1.agent.send,
+    input.data,
+  );
+  if (!sent.ok) return { exitCode: exitCodes.network, ...failureFields(sent) };
+
+  ctx.ui.output(
+    sent.value,
+    agentMessages.started(sent.value.sessionId, sent.value.url),
+  );
+  await deps.store.rememberSession(sent.value.sessionId).catch(() => undefined);
```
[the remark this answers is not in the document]
chajac: Fixed in 143a08356. A failed save now warns with the session id and says to pass --session.

src/commands/agent/send.register.ts
```diff
@@ -0,0 +1,76 @@
...
+  session?: string;
+  timeout: string;
+  workspaceId?: string;
+};
+
+export function registerAgentSendCommand(
+  agent: Command,
+  signals: SignalRegistry,
+): void {
+  declareCommandKind(agent.command("send <message>"), "write")
```
theonly1me, line 31: Looks like we're dropping `--file-paths`,  without that I don't think claude or codex can upload files to Tester, @matt-teran recently implemented this.

src/domains/agent/send.ts
```diff
@@ -0,0 +1,117 @@
...
+  let repliesAtAnswer: number | undefined;
+  if (options.follow && options.session !== undefined) {
+    const before = await ctx.platformClient.callPublicApi(
+      publicContractsV1.agent.get,
+      { sessionId: options.session },
+    );
+    if (!before.ok) {
+      return { exitCode: exitCodes.network, ...failureFields(before) };
+    }
+    repliesAtAnswer = before.value.replies.length;
```
theonly1me: A failed read here returns before the send, so `--follow` can possibly turn a message that would have gone through into one that never leaves.

```suggestion
    if (before.ok) repliesAtAnswer = before.value.replies.length;
```

src/commands/agent/send.register.ts
chajac: Ah, I can add those in thanks for the eyes. Seems like it's not quite in on platform yet so I can pull it over when it is.

src/domains/agent/send.ts
chajac: Fixed in 408d78507.

theonly1me (approved):

PR #1619, opened 2026-09-17
fix(cli): say what a 404 could not find, and which runner

gorangajic:
Every HTTP 404 from the QA Wolf API came back as "could not find X for that environment. Check the --env value" — even when the request had nothing to do with an environment. The two common causes are a runner that is no longer running and a run id from `qawolf runner run` passed to `qawolf run get`. A customer team spent time chasing an environment because of the wording.

Relates to WIZ-12140. Builds on WIZ-12139, which makes the platform name the missing runner in its response; this change prefers that reason when it arrives and stands on its own wording when it does not.

# Overview of Changes

A 404 now answers with what the request named. A runner-targeting command says the runner is not running, names it, and gives the launch command. `qawolf run get` says there is no such run on this team, and that ids printed by `qawolf runner run` belong to that runner — read them with `qawolf runner events run-status`. Only a request genuinely scoped to an environment still points at `--env`.

Runner failures also say which of `--runner`, `QAWOLF_RUNNER_ID` or this directory's stored default chose the id. None of that reached the error before, so a reader of an agent transcript could not tell which runner was meant or why.

These failures exit `8` instead of `4`. Exit `4` reads as "retry", and a terminated runner never comes back, so callers burned their retry budget on an id that could not work. The exit-code doc and the agent skill reference are updated to match.

# Testing

```bash
bun run typecheck
bun run lint
bun run format:check
bun run knip
bun run test
bun run build
```

New unit tests cover each routed 404, including the case where the server still answers a bare "Not found", and the runner-id source on each of the four ways a runner is chosen.

# Checklist

- [x] Self-review completed

A caller branching on exit `4` for a missing runner now sees `8`. That is the point of the change, and the published exit-code contract is updated with it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/shell/platform/describeNotFound.ts
```diff
@@ -0,0 +1,52 @@
...
+export function describeNotFound(
+  subject: NotFoundSubject | undefined,
+  noun: string | undefined,
+  reason: string,
+): PlatformFailure {
+  const explain = (error: string, fallback: string): PlatformFailure => {
+    const detail = reason && !saysNothingMore(reason) ? reason : fallback;
+    return {
+      error,
+      exitCode: exitCodes.notFound,
```
theonly1me, line 48: `run.get` responds with "Run is still being created. Try again in a few seconds." for 404, then we print that under "QA Wolf has no run X on this team", and the `docs/exit-codes.md` says that status will never clear by waiting, so the message is in accurate. Since server is already giving us a meaningful reason, can we use that instead of this one?

src/shell/platform/describeNotFound.ts
```diff
@@ -0,0 +1,52 @@
...
+      return explain(
+        m.notFound404Run(subject.runId),
+        m.runIdMayBeRunnerLocal(subject.runId),
+      );
+    case "other":
+      return explain(m.notFound404(noun), "");
+    // A caller that named no subject is one of the environment-scoped reads:
+    // the flow bundle, an environment's variables, the team's storage.
+    case "environment":
+    case undefined:
```
theonly1me: `team.listStorageFiles` takes `{ teamId, excludePrefixes, pageToken }` and no environment, but `teamStorage.ts` calls `describeRequestError` without a subject, so its 404 still lands here and tells the caller to check `--env`.

I think we should make the `notFound` parameter required so every caller has to name its subject, and pass `{ kind: "other" }` there.

src/shell/platform/describeNotFound.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const m = authMessages.errors.request;
+
+/**
+ * A server reason that only repeats what the status already said. Dropping it
+ * leaves room for wording a reader can act on, and a server that names the
+ * missing thing instead (WIZ-12139) is preferred over anything invented here.
+ */
+const saysNothingMore = (reason: string): boolean =>
+  /^not ?found\.?$/i.test(reason.trim());
```
theonly1me: I think Apex sends `Trigger not found`, `Issue not found`, `Email not found` and `Run not found` etc:

```suggestion
  /^(\w+ )?not ?found\.?$/i.test(reason.trim());
```

src/runnerSdk/givenRunner.ts
```diff
@@ -0,0 +1,10 @@
+import type { TargetedRunner } from "~/domains/interactiveRunner/resolveRunner.js";
+
+/**
+ * An SDK caller names the runner in the call itself, so there is no flag,
+ * variable or stored default behind it for a failure to report.
+ */
+export const givenRunner = (runnerId: string): TargetedRunner => ({
```
theonly1me, line 7: The `SdkResult` is `{ error: string; ok: false }` and `runVerbs` and `lifecycleVerbs` both are returning `read.error` only.

I think we should either carry `errorBody` on `SdkResult` or drop the `given` variant, right now `runnerRequestFailure.test.ts` is the only thing reading it.

theonly1me (approved):

src/shell/platform/describeNotFound.ts
gorangajic: 92f0b2f9

src/shell/platform/describeNotFound.ts
gorangajic: 92f0b2f9

src/runnerSdk/givenRunner.ts
gorangajic: c9d61b99

